PR #1, opened 2019-07-26
Create initial states to setup NW environment

arbulu89:
Initial Netweaver configuration based in: https://www.suse.com/documentation/suse-best-practices/pdfdoc/SAP_HA740_SetupGuide_color_en.pdf

It covers the chapter 3.
- Setup shared disk for ASCS and ERS
- Add entries to /etc/hosts
- Mount sap installation media and sapmnt nfs shares
- Mount needed disks in folder

Disclaimer:
- By now, the states are really hardcoded. We will see how to make it more generic (specially the disk partition)
- Database (and 3rd node in general) configuration has been excluded, so only ASCS and ERS applications setup has been done

### DIFF
--- a/netweaver/setup/shared_disk.sls
+++ b/netweaver/setup/shared_disk.sls
@@ -1,0 +1,52 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+
+{% for node in netweaver.nodes if host == node.host and node.shared_disk_dev is defined %}
+
+{% if node.init_shared_disk is defined and node.init_shared_disk == True %}
+
+label_partition:
+  cmd.run:
+    - name: /usr/sbin/parted -s {{ node.shared_disk_dev }} mklabel gpt
+
+sbd_partition:
+  cmd.run:
+    - name: /usr/sbin/parted -s {{ node.shared_disk_dev }} mkpart primary 1049k 8388k
+    - require:
+      - label_partition
+
+first_partition:
+  cmd.run:
+    - name: /usr/sbin/parted -s {{ node.shared_disk_dev }} mkpart primary 8389k 10.7G
+    - require:
+      - label_partition
+
+second_partition:
+  cmd.run:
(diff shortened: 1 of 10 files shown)


netweaver/setup/shared_disk.sls
```diff
@@ -0,0 +1,52 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+
+{% for node in netweaver.nodes if host == node.host and node.shared_disk_dev is defined %}
+
+{% if node.init_shared_disk is defined and node.init_shared_disk == True %}
+
+label_partition:
+  cmd.run:
+    - name: /usr/sbin/parted -s {{ node.shared_disk_dev }} mklabel gpt
```
diegoakechi, line 10: @arbulu89 AFAIK there is a salt state for that, so we could make it idempotent. I am ok to change it in future PR though.

pillar.example
```diff
@@ -0,0 +1,20 @@
+netweaver:
+  version: 7.5
+  sid: HA1
+  virtual_addresses:
+    192.168.201.111: hacert01
+    192.168.201.112: hacert02
+    192.168.201.115: sapha1as
+    192.168.201.116: sapha1er
+  sapcd_media: your_nfs_share_SWPM_folder
```
diegoakechi: I think `sapcd` is a old reference of a real SAP CD that was distributed. I would change it to something like `swpm_media`

diegoakechi (approved):

MalloZup (approved):
lgtm from my side. Maybe we should create some minimal doc as follow-up explaining the parameters

arbulu89:
> lgtm from my side. Maybe we should create some minimal doc as follow-up explaining the parameters

This is an early stage of development, we will add a lot of things yet until having even the first working version.

netweaver/setup/shared_disk.sls
arbulu89: Ok, I will have look to that in the future. I prefer to have some initial working version instead of perfection hehe In fact, adapting single states shouldn't affect anything else, so It should be a straight-forward change in the future

PR #4, opened 2019-08-13
Use salt module fns to manage shared_disk states

smpsgill:
Used salt module.run functionality to manage the partitions.
Goal is to have these shared disk states idempotent.
TODO: find better alternative to cmd for the rescan_partition state

### DIFF
--- a/netweaver/setup/shared_disk.sls
+++ b/netweaver/setup/shared_disk.sls
@@ -6,36 +6,54 @@
 {% if node.init_shared_disk is defined and node.init_shared_disk == True %}
 
 label_partition:
-  cmd.run:
-    - name: /usr/sbin/parted -s {{ node.shared_disk_dev }} mklabel gpt
+  module.run:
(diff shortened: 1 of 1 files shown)


arbulu89 (commented):
Hey @Simranpal,
I don't disagree with the changes (actually they are more salt based), but do they provide any idempotency.
As far as I see, the changes wouldn't affect at all to the deployment, it would exactly do the same.

smpsgill:
> Hey @Simranpal,
> I don't disagree with the changes (actually they are more salt based), but do they provide any idempotency.
> As far as I see, the changes wouldn't affect at all to the deployment, it would exactly do the same.

Hi @arbulu89 yes that is correct, they will not provide idempotency. I could not find salt state functionalities for managing partitions.  In this case we have to create state in python then?

arbulu89 (approved):
We can merge this changes. So far they look good.
@Simranpal, Please, at least test them and if they work merge yourself

PR #6, opened 2019-08-30
Create a swap space for installing NW

smpsgill:
Initial version TODO:
- Create idempotent states
- Check and add better conditions based on the requirements of SAP installation(NW, HANA and so on) and if each host needs to have the swap space

### DIFF
--- a/netweaver/setup/swap_space.sls
+++ b/netweaver/setup/swap_space.sls
@@ -3,7 +3,7 @@
 
 {% if netweaver.create_swap|default(true) %}
 {% set swap_dir = netweaver.swap_config.directory|default('/var/lib/swap', true) %}
-{% set swap_file =  swap_dir~'/swapfile' %}
+{% set swap_file = swap_dir~'/swapfile' %}
 {% set swap_size = netweaver.swap_config.size|default(128, true) %}
 
 create_swap_dir_{{ host }}:
--- a/netweaver/setup/swap_space.sls
+++ b/netweaver/setup/swap_space.sls
@@ -1,37 +1,38 @@
 {%- from "netweaver/map.jinja" import netweaver with context -%}
 {% set host = grains['host'] %}
 
-{% for node in netweaver.nodes if host == node.host %}
+{% if netweaver.create_swap|default(true) %}
+{% set swap_dir = netweaver.swap_config.directory|default('/var/lib/swap', true) %}
+{% set swap_file =  swap_dir~'/swapfile' %}
+{% set swap_size = netweaver.swap_config.size|default(128, true) %}
 
-{% set swap_file = '/var/lib/swap/swapfile' %}
+create_swap_dir_{{ host }}:
+  file.directory:
+  - name: {{ swap_dir }}
 
-create_swap_dir:
+create_swap_file_{{ host }}:
   cmd.run:
-  - name: mkdir -p /var/lib/swap
-
-create_swap_file:
-  cmd.run:
-  - name: 'dd if=/dev/zero of={{ swap_file }} bs=1048576 count={{ grains["mem_total"] * 2 }} && chmod 0600 {{ swap_file }}'
+  - name: 'dd if=/dev/zero of={{ swap_file }} bs=1048576 count={{ swap_size }} && chmod 0600 {{ swap_file }}'
   - creates: {{ swap_file }}
 
-set_swap_file:
+set_swap_file_{{ host }}:
   cmd.wait:
   - name: 'mkswap {{ swap_file }}'
   - watch:
-    - create_swap_file
+    - create_swap_file_{{ host }}
 
-set_swap_file_status:
+set_swap_file_status_{{ host }}:
   cmd.run:
   - name: 'swapon {{ swap_file }}'
   - unless: grep {{ swap_file }} /proc/swaps
   - require:
-    - set_swap_file
+    - set_swap_file_{{ host }}
 
-{{ swap_file }}:
+mount_swap_{{ host }}:
   mount.swap:
+  - name : {{ swap_file }}
   - persist: True
   - require:
-    - set_swap_file
+    - set_swap_file_{{ host }}
 
 {% endif %}
-{% endfor %}
--- a/netweaver/setup/swap_space.sls
+++ b/netweaver/setup/swap_space.sls
@@ -1,0 +1,37 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+
+{% for node in netweaver.nodes if host == node.host %}
+
+{% set swap_file = '/var/lib/swap/swapfile' %}
+
+create_swap_dir:
+  cmd.run:
+  - name: mkdir -p /var/lib/swap
+
+create_swap_file:
+  cmd.run:
+  - name: 'dd if=/dev/zero of={{ swap_file }} bs=1048576 count={{ grains["mem_total"] * 2 }} && chmod 0600 {{ swap_file }}'
+  - creates: {{ swap_file }}
+
+set_swap_file:
+  cmd.wait:
+  - name: 'mkswap {{ swap_file }}'
+  - watch:
+    - create_swap_file
+
+set_swap_file_status:
+  cmd.run:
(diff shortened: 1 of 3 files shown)


netweaver/setup/swap_space.sls
```diff
@@ -0,0 +1,37 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+
+{% for node in netweaver.nodes if host == node.host %}
```
arbulu89: I would say this for loop is not needed as this operation must be applied to all nodes (actually, this is what will happen with the current implementation, but it's more confusing)

netweaver/setup/swap_space.sls
```diff
@@ -0,0 +1,37 @@
...
+
+{% set swap_file = '/var/lib/swap/swapfile' %}
```
arbulu89: Should we add this as a variable in the pillar file? And put this value as default.
Just wondering if this makes sense or not.
Otherwise this hardcoded value looks good

netweaver/setup/swap_space.sls
```diff
@@ -0,0 +1,37 @@
...
+
+create_swap_dir:
```
arbulu89: We should use the `file.directory` state to make this idempotent instead of using `cmd.run`

netweaver/setup/swap_space.sls
```diff
@@ -0,0 +1,37 @@
...
+  cmd.run:
+  - name: mkdir -p /var/lib/swap
+
+create_swap_file:
```
arbulu89: We should not run the command if the file is already there. Add and `unless/onlyIf` option there checking if the file exist

arbulu89 (changes requested):
In general I like the implementation.
Some comments added.
Maybe we should add a variable to the end of the state names just to avoid potential issues with duplicated names. As example:
https://github.com/salt-formulas/salt-formula-linux/blob/master/linux/storage/swap.sls

Some more suggestions:
- Maybe we should give the option to enable/disable swap creation via pillar. Something like `create_swap: True` being True by default.
- Give the option to select the swap file and size. I don't understand this mem_total * 2 approach. We should have a default value, but provide the option to select size too.

netweaver/setup/swap_space.sls
smpsgill: Yes, the creates will check this. It will Only run if the file or files specified by creates do not exist.

smpsgill:
Hi @arbulu89 I made improvements as per suggestion.
I am also thinking that this state of creating swap should go into our future SAP formula to setup the SAP infrastructure.

netweaver/setup/swap_space.sls
```diff
@@ -0,0 +1,38 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+
+{% if netweaver.create_swap|default(true) %}
+{% set swap_dir = netweaver.swap_config.directory|default('/var/lib/swap', true) %}
+{% set swap_file =  swap_dir~'/swapfile' %}
```
arbulu89: Double space after `=`

pillar.example
```diff
@@ -9,7 +9,11 @@ netweaver:
   sapmnt_inst_media: your_nfs_share_SID_folder
   swpm_folder: your_swpm_folder_relative_to_swpm_media
   sapexe_folder: your_sapexe_sar_folder_relative_to_swpm_media
-
+  create_swap: True
```
arbulu89: What do you think if we just skip the `create_swap: True` and put directly and replace the `swap_config` by that parameter.
Just to remove one dependant parameter. The final result would be:
```
create_swap:
      directory: /var/lib/swap
      size: 128
```
I think this would be more robust.

pillar.example
```diff
@@ -9,7 +9,11 @@ netweaver:
...
+  swap_config:
+      directory: /var/lib/swap
+      size: 128
```
arbulu89: We should add a comment here. Which is the unit of the size?

arbulu89 (changes requested):
Some minor comments here.
Besides that the implementation is good and it's ready to merge

pillar.example
arbulu89: Please, if you add these changes document them in the pillar.example file. Say how the parameters have to be used

netweaver/setup/swap_space.sls
```diff
@@ -0,0 +1,38 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+
+{% if netweaver.create_swap|default(true) %}
```
arbulu89: @Simranpal, you forgot to update the code after updating the pillar example here

pillar.example
```diff
@@ -9,7 +9,10 @@ netweaver:
   sapmnt_inst_media: your_nfs_share_SID_folder
   swpm_folder: your_swpm_folder_relative_to_swpm_media
   sapexe_folder: your_sapexe_sar_folder_relative_to_swpm_media
-
+  swap_config:
```
arbulu89: Please add a comment saying how to use the `swap_config` option. Basically, if this entry exist, the swap partition will be created using the swap configuration parameters.

arbulu89 (changes requested):
@Simranpal, you forget to make some changes

netweaver/setup/swap_space.sls
```diff
@@ -0,0 +1,38 @@
...
+{% set swap_dir = netweaver.swap_config.directory|default('/var/lib/swap', true) %}
+{% set swap_file = swap_dir~'/swapfile' %}
+{% set swap_size = netweaver.swap_config.size|default(128, true) %}
```
arbulu89: I would put 2048 by default, as this is the requited value by SAP

netweaver/setup/swap_space.sls
smpsgill: Hi @arbulu89 yes the existing logic did not work as such for all cases, I have updated it to consider more conditions

smpsgill:
Hi @arbulu89 I tested these changes with various conditions provided in pillar. 
Let me know what you think of the additional logic.

arbulu89 (changes requested):
@Simranpal,
I don't understand the last changes... It was already good in the last commit (with just a minimal change).

All this logic is super confusing in my opinion. We have added extra complexity with this new if/else statement.

Let me suggest other solution.

If we want to enable the swap creation by default (that might be a good idea), let's add this entry to the [defaults.yaml file](https://github.com/SUSE/sapnwbootstrap-formula/blob/master/netweaver/defaults.yaml). There we can add the default `swap_create` data after netweaver (actually, you can even remove the `version` entry, it's not used.

And then just check if the `create_swap` is defined and it's not False.

Something like:

```
# If create_swap entry exists and it's not False
{% if netweaver.create_swap is defined and netweaver.create_swap %}
{% set swap_dir = netweaver.create_swap.directory %}
{% set swap_size = netweaver.create_swap.size %}
# The rest of the state file
```

smpsgill:
> @Simranpal,
> I don't understand the last changes... It was already good in the last commit (with just a minimal change).
> 
> All this logic is super confusing in my opinion. We have added extra complexity with this new if/else statement.
> 
> Let me suggest other solution.
> 
> If we want to enable the swap creation by default (that might be a good idea), let's add this entry to the [defaults.yaml file](https://github.com/SUSE/sapnwbootstrap-formula/blob/master/netweaver/defaults.yaml). There we can add the default `swap_create` data after netweaver (actually, you can even remove the `version` entry, it's not used.
> 
> And then just check if the `create_swap` is defined and it's not False.
> 
> Something like:
> 
> ```
> # If create_swap entry exists and it's not False
> {% if netweaver.create_swap is defined and netweaver.create_swap %}
> {% set swap_dir = netweaver.create_swap.directory %}
> {% set swap_size = netweaver.create_swap.size %}
> # The rest of the state file
> ```

Hi @arbulu89 you are correct. I think I overcomplicated the logic a bit. I will revert to previous logic and you suggestion. Thank you!

arbulu89 (approved):
Thank you for the changes.
I think is ready to be merged now

PR #9, opened 2019-09-17
Add swap space creation

smpsgill:
Add swap state to init.sls


arbulu89:
@Simranpal Have you finally tested this change? If it works properly, let me know and we will merge it.

smpsgill:
Hi @arbulu89 yes I tested this and swap state is applied correctly.

I was not able to do the full deployment because of other error "Node /sapmnt/HA1/profile does not exist." But the changes from PR related to swap work as expected.

arbulu89 (approved):
@Simranpal , Ok when can merge then

PR #10, opened 2019-09-18
Feature/db pas instances

arbulu89:
Add new salt states to install Netweaver DB, PAS and AAS instances.
They follow mostly the same structure than ASCS and ERS.

### DIFF
--- a/netweaver/defaults.yaml
+++ b/netweaver/defaults.yaml
@@ -1,4 +1,6 @@
 netweaver:
+  installation_folder: /tmp/unattended
+  additional_dvds: []
   # Set this entry to false to disable creating swap
   create_swap:
       # Path containing the swapfile
--- a/netweaver/defaults.yaml
+++ b/netweaver/defaults.yaml
@@ -1,6 +1,6 @@
 netweaver:
   clean_nfs: True
-  installation_folder: /tmp/unattended
+  installation_folder: /tmp/swpm_unattended
   additional_dvds: []
   # Set this entry to false to disable creating swap
   create_swap:
--- a/netweaver/defaults.yaml
+++ b/netweaver/defaults.yaml
@@ -1,4 +1,5 @@
 netweaver:
+  clean_nfs: True
   installation_folder: /tmp/unattended
   additional_dvds: []
   # Set this entry to false to disable creating swap
--- a/netweaver/install_aas.sls
+++ b/netweaver/install_aas.sls
@@ -1,22 +1,26 @@
 {%- from "netweaver/map.jinja" import netweaver with context -%}
 {% set host = grains['host'] %}
 
-{% for node in netweaver.nodes if node.host == host and node.sap_instance == 'ascs' %}
+{% for node in netweaver.nodes if node.host == host and node.sap_instance == 'aas' %}
 
 {% set instance = '{:0>2}'.format(node.instance) %}
 {% set instance_name =  node.sid~'_'~instance %}
 
-create_ascs_inifile_{{ instance_name }}:
+create_aas_inifile_{{ instance_name }}:
   file.managed:
-    - source: salt://netweaver/templates/ascs.inifile.params.j2
-    - name: /tmp/ascs.inifile.params
+    - source: salt://netweaver/templates/aas.inifile.params.j2
+    - name: /tmp/aas.inifile.params
     - template: jinja
-    - context: # set up context for template ascs.inifile.params.j2
+    - context: # set up context for template aas.inifile.params.j2
         master_password: {{ node.master_password }}
         sid: {{ node.sid }}
         instance: {{ instance }}
         virtual_hostname: {{ node.virtual_host }}
         download_basket: /swpm/{{ netweaver.sapexe_folder }}
+        schema_name: {{ netweaver.schema_password|default('SAPABAP1') }}
+        schema_password: {{ netweaver.schema_password }}
+        hana_password: {{ netweaver.hana.password }}
+        hana_inst: {{ netweaver.hana.instance }}
 
 netweaver_install_{{ instance_name }}:
   netweaver.installed:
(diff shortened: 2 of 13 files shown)


netweaver/setup/sap_nfs.sls
```diff
@@ -26,7 +26,7 @@ mount_usersapsys_{{ instance_name }}:
     - opts:
       - defaults
 
-{% if netweaver.clean_nfs|default(true) %}
+{% if netweaver.clean_nfs and node.sap_instance == 'ascs' %}
```
smpsgill, line 29: Hi @arbulu89 what if the ERS install starts first, we need to still cleanup in that case first? and during ASCS install, we may still clean resources created by ERS

pillar.example
```diff
@@ -31,4 +55,23 @@ netweaver:
...
+      sap_instance: pas
+
+    - host: hacert04
+      virtual_host: sapha1aas
+      sid: HA1
+      instance: 01
+      root_user: root
+      root_password: linux
+      master_password: your_password
+      sap_instance: pas
```
smpsgill: @arbulu89 is this 'sap_instance' entry not for the aas instance, there is already pas entry above? maybe a typo

templates/aas.inifile.params.j2
```diff
@@ -0,0 +1,139 @@
...
+NW_Delete_Sapinst_Users.removeUsers = true
+
+# Master password
+NW_GetMasterPassword.masterPwd = {{ master_password }}
+
+# Install the SAP HANA database client in a central or local directory. Possible values are: 'SAPCPE', 'LOCAL'
+# NW_HDB_DBClient.clientPathStrategy = LOCAL
+
+# Password of user 'SYSTEM' within the 'SystemDB' tenant in an SAP HANA MultiDB server
+NW_HDB_getDBInfo.systemDbPassword = {{ hana_password }}
```
smpsgill, line 68: It reminds me we may need to consider HANA 1.0 specific parameters in inifile.params, if we are using HANA 1.0 db instance

netweaver/install_aas.sls
```diff
@@ -0,0 +1,62 @@
...
+        hana_password: {{ netweaver.hana.password }}
+        hana_inst: {{ hana_instance }}
+
+wait_for_db_{{ instance_name }}:
+  hana.available:
+    - name: {{ netweaver.hana.host }}
+    - port: 3{{ hana_instance }}15
+    - user: {{ netweaver.schema.name|default('SAPABAP1') }}
+    - password: {{ netweaver.schema.password }}
+    - timeout: 5000
```
smpsgill, line 32: Do we need to wait this long for the hana install to finish (in case hana installation starts in parallel).
Also considering the case where hana installation fails..

pillar.example
```diff
@@ -2,14 +2,38 @@ netweaver:
...
+  # True by default
+  clean_nfs: True
   swpm_media: your_nfs_share_with_swpm_and_sapexe_folders
   sapmnt_inst_media: your_nfs_share_SID_folder
   swpm_folder: your_swpm_folder_relative_to_swpm_media
   sapexe_folder: your_sapexe_sar_folder_relative_to_swpm_media
+  # Folder where the installation files are stored. /tmp/unattended by default. Set None to use
+  # SAP default folders (it will only work with ASCS and ERS)
+  installation_folder: /tmp/unattended
+  # DB/PAS/AAS instances require additional DVD folders like NW Export or HDB Client folder
```
smpsgill, line 19: Hi @arbulu89 do these folders contain some installation files/ or any other files we need to provide? I did not understand whats the purpose of these DVDs

smpsgill (commented):
Hi @arbulu89 , I have added some comments.
I did not test it with the deployment yet, as I was facing other errors with the deployment. I will do it once those are resolved.
And one question regarding the sequence of installing the various instances, Should we enforce certain sequence atleast for installing instances of DB, AAS, PAS?

arbulu89:
> Hi @arbulu89 , I have added some comments.
> I did not test it with the deployment yet, as I was facing other errors with the deployment. I will do it once those are resolved.
> And one question regarding the sequence of installing the various instances, Should we enforce certain sequence atleast for installing instances of DB, AAS, PAS?

About enforcing the instances installation, here the reasons why I have implemented the whole installation this way:
If we want to add this kind of logic to salt states that are being executed in different nodes we would need to add orchestration. This would require a whole update of all our design, as salt master would be required. 
Actually, the current implementation does follow a sequence. If you look more carefully, you will find some states to wait until some data is available in the HANA database `hana.wait_for_connection`.
This modules will perform connections in the database until finally this is possible. In order to start the DB installation, just the HANA has to be up and running. For the other 2 they will wait until some specific schema is installed in the database, that will be created during the DB installation.
Finally is true that PAS and AAS will be executed in parallel, but `sapinst` will manage this by its own. That's why I added the retries. `sapinst` only will work if the conditions are satisfied (for instance the AAS needs the PAS working) so at the end everything will be done in a particular order in all of the executions.

netweaver/setup/sap_nfs.sls
arbulu89: This is something complex yes. The things is that the only the ASCS will install files in this folder at first, as all of the other instances (ERS, DB, PAS, AAS) will check there to find any ASCS instance to start their own installation. So, basically, if ASCS is not installed, all the rest instances cannot do anything. That's why I just set ASCS there. 
Anyway, I'm not totally happy with this condition neither. And eventually we might need to improve the whole setup of Netweaver landscape.

pillar.example
arbulu89: You are right, this is typo. The last one should be `aas`

templates/aas.inifile.params.j2
arbulu89: As we don't even have any HANA 1.0 deployment implementation done, I have not considered this conditions. But yes, in the future we will need to adapt that.
Thanks for the reminder!

netweaver/install_aas.sls
arbulu89: This wait condition doesn't wait just for HANA, it waits until the SAPABAP1 user (or the schema defined in our pillar files) is created in the database. This is done during the DB installation, and this installation takes a long time. That's why I added this big timeout. Anyway, this value doesn't have any meaning, I just added a big value...

pillar.example
arbulu89: Yes. Some Netweaver instances require some additional software besides Netweaver installation files. For example the Netweaver Export and hdbclient installation folders.
If you run the installation manually it will be the same.

netweaver/defaults.yaml
```diff
@@ -1,7 +1,10 @@
 netweaver:
+  clean_nfs: True
+  installation_folder: /tmp/unattended
```
diegoakechi: @arbulu89 Maybe this name is too generic that could collide with an existing folder. What if `/tmp/swpm_uanttended`

pillar.example
diegoakechi: @arbulu89 will you submit for that?

diegoakechi (approved):
LGTM. Two small comments, but they can be addressed in a separate PR.

netweaver/defaults.yaml
MalloZup: `/tmp/netweaver_unattended`?

MalloZup (approved):
looks good. However I miss the functional part of netweaver.
We might also think to create some highlevel doc for usage.

anyways outside of this PR scope thx!

PR #13, opened 2019-10-29
Implement validation of NW formula

smpsgill:
Following the examples of CI validation in other formulas, I create this script.
Please let me know of improvements


arbulu89:
@Simranpal I good evidence that the CI is working is just to enable it and check the result. Could you do that first?

Besides that:
- I wouldn't create a new example folder for this top.sls files. We can put the them in the CI project as they are only used by the CI

smpsgill:
Hi @arbulu89 I corrected the issue which was due to a small typo :/
If you have other suggestions, please let me know, Otherwise I can merge this PR

.travis.yml
```diff
@@ -2,12 +2,27 @@ sudo: required
...
   - docker
 
+addons:
+  apt:
+    sources:
+      - sourceline: 'deb http://repo.saltstack.com/apt/ubuntu/14.04/amd64/latest trusty main'
+        key_url: 'https://repo.saltstack.com/apt/ubuntu/14.04/amd64/latest/SALTSTACK-GPG-KEY.pub'
+    packages:
+        - salt-common
+
```
arbulu89, line 12: We should add the validation stage in `stages` list (is not mandatory anyway)

ci/validate-formula.sh
```diff
@@ -0,0 +1,80 @@
...
+echo " Using host hacert01                       "
+echo "==========================================="
+
+cp pillar.example ci/pillar/netweaver.sls
+cp ci/salt/top.sls .
+
+cat >grains <<EOF
+host: hacert01
+virtual_host: sapha1as
+sap_instance: ascs
```
arbulu89: Is this entry in the `grains` file used for something? I don't think so, we could delete it in that case (the same for all entries)

arbulu89 (approved):
Perfect! 
I think is ready to merge.
I have added some minor comments if you want to polish a bit.
Feel free to merge as soon as you can!

ci/validate-formula.sh
smpsgill: Hi @arbulu89 yes these entries are not needed. I remove these

arbulu89:
@Simranpal I personally wouldn't update the spec/changes file for this PR. The version is upgraded when we want to release a new version of the package and as this PR is not updating any delivered code, it doesn't make sense as the user is not going to notice any difference.

As a summary, we only upgrade version when changes involving the pkg are done, not changes related the project.

smpsgill:
> only upgrade version when changes involvin

Thanks @arbulu89 I understand now :) I will revert those changes to spec/change files.

PR #15, opened 2019-11-02
Add check if sap profile directory exists 

smpsgill:
Check if the directory `/sapmnt/{SID}/profile` exists.
This check is to be done before install of ERS, DB, AAS, PAS instances

Using the salt state fn `file.exists` which will check if the path exists in system.
Please also see if the attempts/intervals are optimal.

### DIFF
--- a/netweaver/install_db.sls
+++ b/netweaver/install_db.sls
@@ -23,6 +23,13 @@ create_db_inifile_{{ instance_name }}:
         hana_password: {{ netweaver.hana.password }}
         hana_inst: {{ hana_instance }}
 
+check_sapprofile_directory_exists_{{ instance_name }}:
+  file.exists:
+    - name: /sapmnt/{{ node.sid.upper() }}/profile
+    - retry:
+        attempts: 10
+        interval: 180
+
 wait_for_hana_{{ instance_name }}:
   hana.available:
     - name: {{ netweaver.hana.host }}
@@ -49,6 +56,7 @@ netweaver_install_{{ instance_name }}:
     - additional_dvds: {{ netweaver.additional_dvds }}
     - require:
       - create_db_inifile_{{ instance_name }}
+      - check_sapprofile_directory_exists_{{ instance_name }}
       - wait_for_hana_{{ instance_name }}
     - retry:
         attempts: {{ node.attempts|default(5) }}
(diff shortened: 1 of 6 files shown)


netweaver/install_db.sls
```diff
@@ -49,6 +56,7 @@ netweaver_install_{{ instance_name }}:
     - additional_dvds: {{ netweaver.additional_dvds }}
     - require:
       - create_db_inifile_{{ instance_name }}
+      - check_sapprofile_directory_exists_{{ instance_name }}
```
arbulu89: Maybe we should put this `require` condition in the previous `wait_for_hana` state. To avoid running this super long check if previous states failed already.

arbulu89 (changes requested):
@Simranpal It looks good! Anyway, we need to fix some details:

- Increase version (0.1.2 ?) and changes files.
- Play with the retry values. It looks quite fine for me (I would personally put smaller interval (30s) and more attempts as this check is really cheap).
- Test, test and test. Please, make sure that everything works (SLE15 would be enough).

arbulu89:
@Simranpal Any update?

netweaver/install_db.sls
smpsgill: I think `wait_for_hana` and `check_sapprofile_directory_exists` are not dependent, logically as well. But creating the dependency will not harm, and may help reduce the long check as you mentioned. SO i will update the `require` condition for `wait_for_hana`

smpsgill:
Hi @arbulu89 I did the changes. Does the spec file and changelog look fine? I have not updated with `osc vc` before and I tried it now.
Otherwise, I would like to test this with full deployment. I could not do it yet as one of packages(python-shaptools) build was failing for me for SLE15. I will put WIP for now, and update after testing it fully.

sapnwbootstrap-formula.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Wed Nov  6 00:03:44 UTC 2019 - Simranpal Singh <simranpal.singh@suse.com>
+
+- Version bump 0.1.2
+  * Add check if sap profile directory exists before NW components
+    install
+  * Change the retry/interval values for profile directory check
```
arbulu89, line 7: This last line wouldn't be needed hehe. Just the changes for the final PR code.

arbulu89 (approved):
It looks fine now!
Feel free to merge as soon as you finish testing!

smpsgill:
Hi @arbulu89 I tested with the deployment and it works fine. The check finished successfully after ~6 retries

PR #17, opened 2019-11-13
Create individual instance passwords and set master password as default value

smpsgill:
Provides the option to set individual passwords for each NW instance and have the Master password as default.

### DIFF
--- a/netweaver/ha_cluster.sls
+++ b/netweaver/ha_cluster.sls
@@ -22,7 +22,7 @@ wait_until_systems_installed:
     - dispstatus: GREEN
     - sid: {{ node.sid.lower() }}
     - inst: {{ instance }}
-    - password: {{ node.master_password }}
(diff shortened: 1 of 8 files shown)


smpsgill:
I will test it with NW deployment and update

Simranpal:
As we are installing the NW components on different hosts and if we set individual passwords , it may not match the SAP requirements below:

"If you use local operating system user accounts instead of central user management (for example, NIS), users <sapsid>adm and sapadm must have the same password on all hosts.""

from Guide:

https://help.sap.com/viewer/910828cec5d14d6685da380aec1dc4ae/CURRENT_VERSION/en-US/f35e144d63224704831653b1eff76ddd.html

smpsgill:
Closing this PR as it is replaced by https://github.com/SUSE/sapnwbootstrap-formula/pull/19 which provides the correct functionality

PR #18, opened 2019-11-14
Feature/create sidadm user

arbulu89:
sidadm creation is needed for the latest NW version as a previous step. It's not mandatory, but it solves some issues if the NW installation folders are created beforehand as we do in this formula.


smpsgill:
LGTM! Thanks

smpsgill (approved):

diegoakechi (approved):
Great work! Thanks @arbulu89

PR #19, opened 2019-11-19
Provide option to set sapadm and sidadm passwords

smpsgill:
This PR provides the option to set individual `sapadm` and `<sid>adm` passwords for each NW instance, having the Master password as default. The idea is to provide the same level of configuration a user can modify from the SAP GUI installer.

### DIFF
--- a/netweaver/install_ascs.sls
+++ b/netweaver/install_ascs.sls
@@ -13,6 +13,8 @@ create_ascs_inifile_{{ instance_name }}:
     - template: jinja
     - context: # set up context for template ascs.inifile.params.j2
         master_password: {{ node.master_password }}
+        sap_adm_password: {{ node.sap_adm_password|default(node.master_password) }}
+        sid_adm_password: {{ node.sid_adm_password|default(node.master_password) }}
         sid: {{ node.sid }}
         instance: {{ instance }}
         virtual_hostname: {{ node.virtual_host }}
--- a/netweaver/install_ascs.sls
+++ b/netweaver/install_ascs.sls
@@ -24,7 +24,7 @@ netweaver_install_{{ instance_name }}:
   netweaver.installed:
     - name: {{ node.sid.lower() }}
     - inst: {{ instance }}
-    - password: {{ netweaver.master_password }}
+    - password: {{ netweaver.sid_adm_password }}
     - software_path: {{ netweaver.swpm_folder }}
     - root_user: {{ node.root_user }}
     - root_password: {{ node.root_password }}
--- a/netweaver/install_ascs.sls
+++ b/netweaver/install_ascs.sls
@@ -12,9 +12,9 @@ create_ascs_inifile_{{ instance_name }}:
     - name: /tmp/ascs.inifile.params
     - template: jinja
     - context: # set up context for template ascs.inifile.params.j2
-        master_password: {{ node.master_password }}
-        sap_adm_password: {{ node.sap_adm_password|default(node.master_password) }}
-        sid_adm_password: {{ node.sid_adm_password|default(node.master_password) }}
+        master_password: {{ netweaver.master_password }}
+        sap_adm_password: {{ netweaver.sap_adm_password|default(netweaver.master_password) }}
+        sid_adm_password: {{ netweaver.sid_adm_password|default(netweaver.master_password) }}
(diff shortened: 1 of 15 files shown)


smpsgill:
Need to identify the other areas, where the `sapadm` and `<sid>adm` passwords could replace the master password, such as enabling HA, sapservices, start/stop NW instances using the sapcontrol would require the `<sid>adm` password

netweaver/install_ascs.sls
```diff
@@ -22,7 +24,7 @@ netweaver_install_{{ instance_name }}:
   netweaver.installed:
     - name: {{ node.sid.lower() }}
     - inst: {{ instance }}
-    - password: {{ node.master_password }}
+    - password: {{ netweaver.master_password }}
```
arbulu89: Is this the `master_password` or the `sidadm` password? I would say the 2nd, but I'm not sure

arbulu89: The same for the other sap instances

arbulu89 (changes requested):
@Simranpal Some comments.
- I guess that you confirmed that the same passwords usage in all hosts is the way to go. If this is the case perfect, otherwise, we need to provide to set individual passwords.
- I don't find the changes in the templates for `ascs`. `pas` and `aas` and they should have the same parameter. Maybe we need to update the template to add this option. Otherwise it doesn't have sense to pass the values.
- Maybe we could improve the next line that is repeated many times:
`netweaver.sap_adm_password|default(netweaver.master_password)`
and put it in other place (`defaults`, `map.jinja`), just to do it once. What do you think?

smpsgill:
> @Simranpal Some comments.
> 
> * I guess that you confirmed that the same passwords usage in all hosts is the way to go. If this is the case perfect, otherwise, we need to provide to set individual passwords.

That's what I have found so far from documentation and search.

> * I don't find the changes in the templates for `ascs`. `pas` and `aas` and they should have the same parameter. Maybe we need to update the template to add this option. Otherwise it doesn't have sense to pass the values.

I noticed that those passwords in those instances templates were commented out. Maybe you changed it or was there when you copied the template?
I can add the interpolation for relevant passwords in those templates as well.

> * Maybe we could improve the next line that is repeated many times:
>   `netweaver.sap_adm_password|default(netweaver.master_password)`
>   and put it in other place (`defaults`, `map.jinja`), just to do it once. What do you think?

That would be good, but for 
1. adding this line in defaults.yaml, I would have to import the password pillar info there which doesnt make sense.
2. I guess you mean adding something like `{% set instance_password = netweaver.sap_adm_password|default(netweaver.master_password) %}
` to map.jinja which would mean importing  this instance_password.
3. Setting this variable in each file 
Is there other efficient way you think?

arbulu89:
Hey @Simranpal ,

> I noticed that those passwords in those instances templates were commented out. Maybe you changed it or was there when you copied the template?
I can add the interpolation for relevant passwords in those templates as well.

I used the templates that are created by the manual execution, so I guess that the `sapinst` tool itself doesn't uncomment this lines if the passwords are the same, but we have to. The passwords must be applied to all instances, so if don't change the template, it won't, right?

About the default password thing, I don't have the answer, I wanted to make you research for a cleaner solution hehe I have not spent time looking options. For me, it it was possible, trying to eliminate these duplicates would make sense.

But this is just an improvement, as far as the logic works, and the passwords are applied as expected I don't really care, we can work on that in a future PR too.

Don't take every suggestion I do like a a `must` hehe There are just observations.

netweaver/install_ers.sls
```diff
@@ -39,7 +41,7 @@ netweaver_install_{{ instance_name }}:
     - product_id: NW_ERS:NW750.HDB.ABAPHA
     - cwd: {{ netweaver.installation_folder }}
     - additional_dvds: {{ netweaver.additional_dvds }}
-    - ascs_password: {{ node.master_password }}
+    - ascs_password: {{ netweaver.master_password }}
```
arbulu89: Shouldn't this be `{{ netweaver.sid_adm_password|default(netweaver.master_password) }}` ?

arbulu89 (approved):
@Simranpal Looks good!
Just a single comment added.
*Upgrade the package version changing `spec` and `changes` file.
Then it's ready to merge

PR #22, opened 2019-12-05
Feature/azure support

arbulu89:
Add azure support.

- Use sockat
- Use nfs share to mount ASCS/ERS shared data

### DIFF
--- a/netweaver/setup/mount.sls
+++ b/netweaver/setup/mount.sls
@@ -25,6 +25,24 @@ mount_{{ node.sap_instance.lower() }}_{{ instance_name }}:
     - opts:
       - defaults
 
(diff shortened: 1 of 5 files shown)


netweaver/setup/mount.sls
```diff
@@ -5,33 +5,47 @@
 {% set instance = '{:0>2}'.format(node.instance) %}
 {% set instance_name =  node.sid~'_'~instance %}
 
-{% if node.sap_instance.lower() == 'ascs' %}
+{% if node.sap_instance.lower() in ['ascs', 'ers'] %}
+
+{% if ':' in node.shared_disk_dev %} # This means that the device is a nfs share
```
diegoakechi, line 10: Not sure about that, but I don't have a better suggestion here...

diegoakechi (approved):

PR #24, opened 2020-01-15
Move the step of update sapservices

smpsgill:
update sapservices state after stoping the instances

### DIFF
--- a/netweaver/ha_cluster.sls
+++ b/netweaver/ha_cluster.sls
@@ -27,13 +27,6 @@ wait_until_systems_installed:
         attempts: 20
         interval: 30
 
-update_sapservices_{{ instance_name }}:
-    netweaver.sapservices_updated:
-      - name: {{ node.sap_instance.lower() }}
-      - sid: {{ node.sid.lower() }}
-      - inst: {{ instance }}
-      - password: {{ netweaver.sid_adm_password|default(netweaver.master_password) }}
-
 stop_sap_instance_{{ instance_name }}:
   module.run:
     - netweaver.execute_sapcontrol:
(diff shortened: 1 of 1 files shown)


arbulu89 (commented):
@Simranpal I don't really know if we can do this. The `sapservices_updated` state uses the sapcontrol command, and if the instance is stopped, the command don't usually work... Maybe it works this time due the timings, but I would like to be sure of this. Just try to stop the service manually and run the `GetSystemInstanceList` command to see if it always work.

Besides that, I'm not sure why this change would affect to the next start process. The unique explanation I can find is that the Stop was failling, and this change fixes that avoid issues in the Start. 

I would like to find at least a explanation why this change fixes the issue to see that is not something random.

smpsgill:
Hi @arbulu89 I suspected that shutdown was not complete, maybe the stop state was not graceful and failed the next Start process. But this is speculation and not concrete. Also as per the guide steps, it had stop stage before updating `'/usr/sap/sapservices'`

And you are correct. I tried with more deployments and it infact seems random. I faced the issue now with one deployment. And I can also confirm that the `GetSystemInstanceList` did not run after I stopped the service manually.

`'/usr/sap/sapservices'` is being used by `sapstartsrv` in general to start/stop instances, by installer,users(via `sapcontrol`) as well as cluster in our case. Now I can only think of maybe adding a wait time before issuing the stop command.
All of the steps we are doing are correct, as per the guide, and it might be an issue with the swpm version maybe.

arbulu89:
@Simranpal After merging https://github.com/SUSE/sapnwbootstrap-formula/pull/26 I think we can close this PR. What do you think?

smpsgill:
> @Simranpal After merging #26 I think we can close this PR. What do you think?

Yes @arbulu89 lets close this

PR #25, opened 2020-01-17
Feature/adapt s4hana

arbulu89:
Adapt the installation states to use the product id as parameter, this way we can install different products (Netweaver, S4HANA, etc)


smpsgill (approved):
Looks good! @arbulu89

PR #26, opened 2020-01-20
Hotfix/fix ascs restart

arbulu89:
This PR fixes the issue of the ASCS restart failure.
Time to time (randomly) the disk partition fails, so after that the SAP instances don't work properly as they depend on this shared disk. This `sleep` executions make the process work properly


smpsgill (approved):
LGTM! Thanks @arbulu89

PR #27, opened 2020-01-23
Add saptune apply solution to netweavere formula

MalloZup:
add saptune to netweaver

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -37,6 +37,11 @@ netweaver:
   # Enable operations in ASCS and ERS to set HA environment correctly (HA cluster is not installed)
   ha_enabled: True
 
+  # saptune solution to apply to all nodes ( by default nothing is applied)
+  # you can also use this to a single node if need to differ. see host hacert02
+  # Warning: only a unique solution can exist into a node.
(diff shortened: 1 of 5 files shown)


MalloZup:
I think this PR is already good but I'm trying to setup netweaver deployement right now to test it

pillar.example
```diff
@@ -37,6 +37,11 @@ netweaver:
   # Enable operations in ASCS and ERS to set HA environment correctly (HA cluster is not installed)
   ha_enabled: True
 
+  # saptune solution to apply to all nodes ( by default nothing is applied)
+  # you can also use this to a single node if need to differ. see hana2 
+  # Warning: only a unique solution can exist into a node.
+  saptune_solution: 'HANA'
```
arbulu89: I would put `NETWEAVER` here as example

arbulu89 (approved):
@MalloZup Looks good! Just the comment about the example.
If you test and works it's ready to go!

pillar.example
MalloZup: done thx

MalloZup:
Ok i have managed to deploy and I could test this works fine. merging

PR #28, opened 2020-01-24
Add the virtual_host_mask parameter to the nodes

arbulu89:
Add virtual_host_mask option to the installation (gcp requires 32 mask)
Related to: https://github.com/SUSE/salt-shaptools/pull/55

PD: I have replaced the default `eth1` to `eth0` as this 2nd is more common scenario


MalloZup (approved):
thx

PR #29, opened 2020-01-28
Feature/gcp resources

arbulu89:
In GCP the `nic` and `netmask=32` are needed. This PR adapts the template to support these paremeters.

The other providers will continue as they work now.

PD: The addition of `-` in all around the template is to avoid the creation of new lines in the created file from the template.


diegoakechi (approved):
LGTM.

PR #32, opened 2020-03-03
Feature/gcp resources

arbulu89:
Use the gcp specific Resource and Fence agents

### DIFF
--- a/sapnwbootstrap-formula.changes
+++ b/sapnwbootstrap-formula.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Mar  3 09:52:45 UTC 2020 - Xabier Arbulu <xarbulu@suse.com>
+
+- Version 0.2.5
+  * Adapt the cluster template to use the proper gcp agents 
+
+-------------------------------------------------------------------
 Fri Feb 21 08:38:47 UTC 2020 - Xabier Arbulu <xarbulu@suse.com>
 
 - Version 0.2.4
--- a/sapnwbootstrap-formula.changes
+++ b/sapnwbootstrap-formula.changes
@@ -2,7 +2,8 @@
 Tue Mar  3 09:52:45 UTC 2020 - Xabier Arbulu <xarbulu@suse.com>
 
 - Version 0.2.5
-  * Adapt the cluster template to use the proper gcp agents 
+  * Adapt the cluster template to use the proper gcp agents
+  (bsc#1161898, bsc#1160933)
 
 -------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


templates/cluster_resources.j2
```diff
@@ -58,17 +58,17 @@ primitive rsc_ip_{{ sid }}_ERS{{ ers_instance }} ocf:suse:aws-vpc-move-ip \
 {%- elif cloud_provider == "google-cloud-platform" %}
 
 # This stonith resource will be duplicated for each node in the cluster
-primitive rsc_gcp_stonith_{{ sid }}_HDB{{ ascs_instance }}_{{ grains['host'] }} stonith:fence_gce \
+primitive rsc_gcp_stonith_{{ sid }}_{{ grains['host'] }} stonith:fence_gce \
     params action=reboot plug={{ grains['host'] }} pcmk_host_map="{{ grains['host'] }}:{{ grains['host'] }}" \
     meta target-role=Started
 
-primitive rsc_ip_{{ sid }}_ASCS{{ instance }} ocf:heartbeat:gcp-vpc-move-route \
+primitive rsc_ip_{{ sid }}_ASCS{{ ascs_instance }} ocf:heartbeat:gcp-vpc-move-route \
```
smpsgill, line 65: Should there be underscore _ before instance numbers . Same with ERS.

arbulu89: @Simranpal We don't put it usually to be honest

sapnwbootstrap-formula.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Mar  3 09:52:45 UTC 2020 - Xabier Arbulu <xarbulu@suse.com>
+
+- Version 0.2.5
```
diegoakechi, line 4: @arbulu89 Don't forget to include the bug number here ;)

diegoakechi (approved):
LGTM. Just a small comment. @arbulu89 Will you introduce other PR to include the other resource agents, like move VIP?

sapnwbootstrap-formula.changes
arbulu89: I think is better to just add the bug trackers when we are going to ship to SLE. I don't think we will ship all the updated versions... So when we have enough worthy changes to be released, we can update them the changes file. What do you think?

About adding more resources, by now we only support the IP management with routes (not alias). This is because the other usage would require different infrastructure creation, and by now, with terraform we only create the routes way. We could implement and give the 2 options in the future, that's true

PR #33, opened 2020-03-05
Add monitoring states for deploy sap_host_exporter

MalloZup:
I will quick test this manually otherwise looks good to me


sapnwbootstrap-formula.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Mar  5 14:52:58 UTC 2020 - Dario Maiocchi <dmaiocchi@suse.com>
+
+- Version 0.2.5
+  * Add pillar for activating sap_host_exporter on nodes
```
arbulu89: I would rewrite by: `Add option to activate the sap_host_exporter on the nodes`

netweaver/init.sls
```diff
@@ -7,3 +9,6 @@ include:
   - netweaver.install_db
   - netweaver.install_pas
   - netweaver.install_aas
+  {% if netweaver.sap_host_exporter is sameas true %}
```
arbulu89: What happens if the entry is not set? Does it work?
I would add the entry in the `defaults.yml`. It will make it easier.

Finally, could you add `-` to the `if` statements.This avoids the creation or new lines in the rendered file, what makes everything cleaner. Like:
`{%- if netweaver.sap_host_exporter is sameas true %}`
And
`{%- endif %}`

arbulu89 (changes requested):
@MalloZup I don't know if it's a good idea to start with this automation.
Now you are enabling the exporter in all the nodes even though all of them are different. I suppose that all host exporters have different `tcp` ports to connect to, and most probably we will need to configure this in each node. For this reason, maybe we will need to enable the exporter individually in each node instead of a generic place.
We can change that in the future. But I would like to have a clear idea about what we are doing and how is the architecture of the exporter before thinking on automation

MalloZup:
@arbulu89  ok thx for review. good point. I was assuming we install the exporter in all Nodes.

the exporter sap_host basically right now has `url` where we can configure the SOAP service port and url where we will do the call and grep the metrics.

Perhaps for this point in time, we can activate the netweaver exporter in the deployment project and split to formula later

MalloZup:
This is already specified in the pillar file in each node, as we need to install it first
You will only need to loop in the nodes, get the instance, update the config file, and them start the exporter
https://github.com/SUSE/sapnwbootstrap-formula/blob/master/pillar.example#L77

    github.com
    SUSE/sapnwbootstrap-formula
    SAP Netweaver bootstrap salt formula. Contribute to SUSE/sapnwbootstrap-formula development by creating an account on GitHub.

For example
https://github.com/SUSE/sapnwbootstrap-formula/blob/master/pillar.example#L90

    github.com
    SUSE/sapnwbootstrap-formula
    SAP Netweaver bootstrap salt formula. Contribute to SUSE/sapnwbootstrap-formula development by creating an account on GitHub.

I'm leaving! See you

netweaver/init.sls
MalloZup: done

MalloZup:
@arbulu89 I think should be ok. I will give it a try manually or with the deployment project

netweaver/monitoring.sls
```diff
@@ -0,0 +1,33 @@
...
+# for each node the instance number change so we need to adapt configuration
+sap_host_exporter_configuration:
+  file.managed:
+    - name: /etc/sap_host_exporter.yaml
+    - contents: |
+         port: "9680"
+         addres: "0.0.0.0"
+         log-level: "info"
+         sap-control-url: "http://127.0.0.1:5{{ node.instance }}13"
+         sap-control-user:  "{{netweaver.sap_host_exporter.sap_control_user}}"
```
arbulu89: Double space after `:`. Just 1 is enough

netweaver/monitoring.sls
```diff
@@ -0,0 +1,33 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+
+{% for node in netweaver.nodes if host == node.host %}
```
arbulu89: Even though it doesn't affect, I will use the `for` loop only in the configuration file creation. The other 2 states can be executed outside the loop without any problem

pillar.example
```diff
@@ -68,6 +68,13 @@ netweaver:
   product_id: NW750.HDB.ABAPHA
   #product_id: S4HANA1709.CORE.HDB.ABAPHA
 
+
+  # optional: enable sap_host_exporter by default disabled
```
arbulu89: Describe that the exporter will be installed and configure in all the nodes

netweaver/defaults.yaml
```diff
@@ -11,3 +11,8 @@ netweaver:
       directory: /var/lib/swap
       # Swap size in MB
       size: 2048
+  # by default the exporter is disabled
+  sap_host_exporter:
+      enabled: false
+      sap_control_user: ""
```
arbulu89: I wouldn't put the user and password with empty default options. I would remove them. They will never be empty strings, so we have to require them. 
In fact, if `enabled` is false they are not used.

arbulu89: In fact, I wouldn't require these 2 options. These parameters are already available in the pillar file in `sid_adm_password` and that `sap_control_user` is created by `{sid}adm` and the sid is available in all nodes pillar file. Use them, the less new parameters the better.

netweaver/monitoring.sls
```diff
@@ -0,0 +1,33 @@
...
+
+prometheus_sap_host_exporter_pkg:
+  pkg.installed:
+    - name: prometheus-sap_host_exporter
+
+# for each node the instance number change so we need to adapt configuration
+sap_host_exporter_configuration:
+  file.managed:
+    - name: /etc/sap_host_exporter.yaml
```
arbulu89: `/etc/sap_host_exporter.yaml`, are you sure?
I would use `/etc/sap_host_expoter/sap_host_exporter.yaml`, if we need to add additional things to the configuration folder.
**Anyway, if you do the same in the `ha_cluster_exporter` this is fine, just to be aligned**

arbulu89 (changes requested):
@MalloZup Some improvements suggested

netweaver/defaults.yaml
MalloZup: ok will check thx

netweaver/monitoring.sls
MalloZup: in ha_cluster exporter we don't need configuration file. (since we don't need to put some special Number which vary from host to host)

I wouldn't have add it  here either :grin: 


regarding the conf file is correct, we use same as ha_cluster_exporter conf path.

netweaver/monitoring.sls
```diff
@@ -0,0 +1,30 @@
...
+{% for node in netweaver.nodes if host == node.host %}
+sap_host_exporter_configuration:
+  file.managed:
+    - name: /etc/sap_host_exporter.yaml
+    - contents: |
+         port: "9680"
+         addres: "0.0.0.0"
+         log-level: "info"
+         sap-control-url: "http://127.0.0.1:5{{ node.instance }}13"
+         sap-control-user: "{{ node.sid.lower() }}"
```
arbulu89: You will need to change this `sap-control-user` line. The user is created as `"{{ node.sid.lower() }}adm"` (check that last `adm`)

netweaver/monitoring.sls
```diff
@@ -0,0 +1,30 @@
...
+sap_host_exporter_exporter_service:
+  service.running:
+    - name: prometheus-sap_host_exporter
+    - enable: True
+    - restart: True
+    - require:
+      - pkg: prometheus_sap_host_exporter_pkg
+      - file: sap_host_exporter_configuration
+
+# for each node the instance number change so we need to adapt configuration
```
arbulu89: `... the instance number change**s**...`

netweaver/monitoring.sls
```diff
@@ -0,0 +1,30 @@
...
+{% for node in netweaver.nodes if host == node.host %}
```
arbulu89, line 18: Where are you getting this `host` from?
I guess you will need to add `{% set host = grains['host'] %}` at the top of the file, right?

netweaver/monitoring.sls
```diff
@@ -0,0 +1,30 @@
...
+sap_host_exporter_configuration:
+  file.managed:
+    - name: /etc/sap_host_exporter.yaml
+    - contents: |
+         port: "9680"
+         addres: "0.0.0.0"
```
arbulu89: It should be `addres**s**`, right? (double `s` at the end)

pillar.example
```diff
@@ -68,6 +68,13 @@ netweaver:
...
+  # the exporter will be installed and configure in all the nodes if activated
```
arbulu89: Please, fix the typos. I think it should be:
```
# optional: enable sap_host_exporter (disabled by default)
# the exporter will be installed and configured in all the nodes if it is enabled
```

sapnwbootstrap-formula.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Mar 10 09:58:57 UTC 2020 - Dario Maiocchi <dmaiocchi@suse.com>
+-  Version 0.2.6
+  * Add option to activate the sap_host_exporter on the nodes
+  * Add user and password pillar for exporter
```
arbulu89: Is this last comment making any sense?

arbulu89 (changes requested):
@MalloZup More changes requested hehe (mostly typos and some other improvements)

PD: Please to follow the same convention in the commit messages please! Start with capital letter always, and try to add a more meaningful messages

netweaver/monitoring.sls
MalloZup: https://github.com/SUSE/sap_host_exporter/pull/12
yep

netweaver/monitoring.sls
MalloZup: yeah always forgetting the s for 3 person :/ :grin:

netweaver/monitoring.sls
MalloZup: done thx

arbulu89 (approved):
Thanks @MalloZup 
Now it looks fine.
In the future, I would add the option to select the port and log level from the formula.

PR #35, opened 2020-03-10
Hotfix/aws stonith action

arbulu89:
Fix to use `off` instead of `poweroff` in the stonith action (`poweroff` is deprecated and it raises a warning message that block crm commands)
In fact, these configuration will only apply to AWS as the official guides say.


MalloZup (approved):
ok thx
!

PR #38, opened 2020-03-11
Feature/update config file

arbulu89:
In order to change the installation process, the most common thing is to update the installation configuration file. We have an initial template with the most important parameters, but we cannot cover everything. 
For that, I have implemented a way to introduce custom changes to the config files.
Basically, a new `extra_parameters` variable, that works as a dictionary where you can put `key/value` entries that will be added/changed in the config files for each node.

Besides that, I have added the option to customize the retries of the SAP instances installation, as for example S4/HANA takes a lot of time (much more than NETWEAVER).

**Pending: Update changes and spec file. I will do that when the PR is approved and ready to go**


netweaver/install_aas.sls
```diff
@@ -9,11 +9,12 @@
 
 {% set product_id = node.product_id|default(netweaver.product_id) %}
 {% set product_id = 'NW_DI:'~product_id if 'NW_DI' not in product_id else product_id %}
+{% set inifile = '/tmp/aas.inifile'~instance_name~'.params' %}
```
smpsgill, line 12: maybe this variable can be name `aas_inifile`

netweaver/install_aas.sls
```diff
@@ -28,6 +29,16 @@ create_aas_inifile_{{ instance_name }}:
         hana_password: {{ netweaver.hana.password }}
         hana_inst: {{ hana_instance }}
 
+{% if node.extra_parameters is defined %}
+update_aas_inifile_{{ instance_name }}:
+  module.run:
+    - netweaver.update_conf_file:
+      - conf_file: {{ inifile }}
+      - {%- for key,value in node.extra_parameters.items() %}
```
smpsgill, line 37: Will the `-` hypen also not come at end, ie before `%}`

netweaver/install_pas.sls
```diff
@@ -9,11 +9,12 @@
 
 {% set product_id = node.product_id|default(netweaver.product_id) %}
 {% set product_id = 'NW_ABAP_CI:'~product_id if 'NW_ABAP_CI' not in product_id else product_id %}
+{% set inifile = '/tmp/pas.inifile'~instance_name~'.params' %}
```
smpsgill, line 12: same here, and for other instances as well the name for variable `inifile` could be changed

smpsgill (changes requested):
Hi @arbulu89 It looks good. I have added some minor comments.

netweaver/install_aas.sls
diegoakechi: As we have the salt files split into each component already, I don't see the need to specialize the variable.

netweaver/install_aas.sls
diegoakechi: LGTM. See: https://jinja.palletsprojects.com/en/master/templates/#whitespace-control

diegoakechi (approved):

netweaver/install_aas.sls
arbulu89: I chose this implementation, because even though it's an strange scenario, we could have more than one `aas` or `pas` in the same host, with different `sid` and `instance number`. Just to make the file name more meaningful and descriptive. I think it's a good practice.

netweaver/install_aas.sls
arbulu89: No, the `-` only comes at the beginning.

netweaver/install_pas.sls
arbulu89: Honestly, as we can use the name we really want, I prefer to have the most descriptive name possible. I don't see any value having a generic name.

netweaver/install_aas.sls
MalloZup: yes we use already this `-` in begin only in other projects

netweaver/install_aas.sls
MalloZup: ok

netweaver/install_aas.sls
MalloZup, line 37: perhaps we could rename `key` to `option`. Option/value sound less technical and intuitive then a generic key/value, since we are adding configuration option.

netweaver/install_ascs.sls
```diff
@@ -23,6 +24,16 @@ create_ascs_inifile_{{ instance_name }}:
         virtual_hostname: {{ node.virtual_host }}
         download_basket: {{ netweaver.sapexe_folder }}
 
+{% if node.extra_parameters is defined %}
+update_ascs_inifile_{{ instance_name }}:
+  module.run:
+    - netweaver.update_conf_file:
+      - conf_file: {{ inifile }}
+      - {%- for key,value in node.extra_parameters.items() %}
```
MalloZup, line 32: same here. maybe option or parameter is better then key

MalloZup (approved):
looks good to me. perhaps we could just rename the `key` variable to `option` or `parameter` which sound more intuitive imho in this context

arbulu89:
@Simranpal Could you check my comments? If so, just approve the PR, otherwise we cannot continue with the merge (once you ask for changes, this same user needs to approve the PR)

smpsgill (approved):
Thanks @arbulu89 Looks good!

PR #40, opened 2020-03-12
Fix non unique ID error for salt state

MalloZup:
fix error: found conflicting ID 'sap_host_exporter_configuration'

### DIFF
--- a/netweaver/monitoring.sls
+++ b/netweaver/monitoring.sls
@@ -1,26 +1,36 @@
 {%- from "netweaver/map.jinja" import netweaver with context -%}
 {% set host = grains['host'] %}
 
+
 prometheus_sap_host_exporter_pkg:
   pkg.installed:
     - name: prometheus-sap_host_exporter
 
+# the sid, instance number pair of a node is unique, so we need to adapt configuration
+{% for node in netweaver.nodes if host == node.host %}
+
+{% set instance = node.instance %}
+{% set instance = '{:0>2}'.format(node.instance) %}
+{% set daemon_instance = '{}_{}'.format(node.sid, instance) %}
+
+# don't install exporter for db instance
+{%- if not node.sap_instance == "db" %}
+
 sap_host_exporter_exporter_service:
   service.running:
-    - name: prometheus-sap_host_exporter
+    - name: prometheus-sap_host_exporter@{{ daemon_instance }}
     - enable: True
     - restart: True
     - require:
       - pkg: prometheus_sap_host_exporter_pkg
-      - file: sap_host_exporter_configuration
+      - file: sap_host_exporter_configuration_{{ daemon_instance }}
 
-# for each node the instance number changes so we need to adapt configuration
-{% for node in netweaver.nodes if host == node.host %}
-sap_host_exporter_configuration:
+# we can have multiple exporters in the same host in that moment, we need to change the port used by the exporter
+sap_host_exporter_configuration_{{ daemon_instance }}:
   file.managed:
-    - name: /etc/sap_host_exporter.yaml
+    - name: /etc/sap_host_exporter{{ daemon_instance }}.yaml
     - contents: |
-         port: "9680"
+         port: "968{{ loop.index }}"
          address: "0.0.0.0"
          log-level: "info"
          sap-control-url: "http://127.0.0.1:5{{ node.instance }}13"
--- a/netweaver/monitoring.sls
+++ b/netweaver/monitoring.sls
@@ -7,11 +7,10 @@ prometheus_sap_host_exporter_pkg:
 
 # the sid, instance number pair of a node is unique, so we need to adapt configuration
 {% for node in netweaver.nodes if host == node.host and node.sap_instance != "db" %}
-{% set instance = node.instance %}
 {% set instance = '{:0>2}'.format(node.instance) %}
 {% set daemon_instance = '{}_{}'.format(node.sid, instance) %}
 
-sap_host_exporter_exporter_service_{{ daemon_instance }}:
+sap_host_exporter_service_{{ daemon_instance }}:
   service.running:
     - name: prometheus-sap_host_exporter@{{ daemon_instance }}
     - enable: True
@@ -26,12 +25,12 @@ sap_host_exporter_exporter_service_{{ daemon_instance }}:
 # we can have multiple exporters in the same host in that moment, we need to change the port used by the exporter
 sap_host_exporter_configuration_{{ daemon_instance }}:
   file.managed:
-    - name: /etc/sap_host_exporter{{ daemon_instance }}.yaml
+    - name: /etc/sap_host_exporter/{{ daemon_instance }}.yaml
     - contents: |
          port: "968{{ loop.index0 }}"
          address: "0.0.0.0"
          log-level: "info"
-         sap-control-url: "http://127.0.0.1:5{{ node.instance }}13"
+         sap-control-url: "http://127.0.0.1:5{{ instance }}13"
          sap-control-user: "{{ node.sid.lower() }}adm"
          sap-control-password: "{{ netweaver.sid_adm_password|default(netweaver.master_password) }}"
     - require:
--- a/netweaver/monitoring.sls
+++ b/netweaver/monitoring.sls
@@ -1,22 +1,17 @@
 {%- from "netweaver/map.jinja" import netweaver with context -%}
 {% set host = grains['host'] %}
 
-
 prometheus_sap_host_exporter_pkg:
   pkg.installed:
     - name: prometheus-sap_host_exporter
 
 # the sid, instance number pair of a node is unique, so we need to adapt configuration
-{% for node in netweaver.nodes if host == node.host %}
-
+{% for node in netweaver.nodes if host == node.host and node.sap_instance != "db" %}
 {% set instance = node.instance %}
 {% set instance = '{:0>2}'.format(node.instance) %}
 {% set daemon_instance = '{}_{}'.format(node.sid, instance) %}
 
-# don't install exporter for db instance
-{%- if not node.sap_instance == "db" %}
-
-sap_host_exporter_exporter_service:
+sap_host_exporter_exporter_service_{{ daemon_instance }}:
   service.running:
     - name: prometheus-sap_host_exporter@{{ daemon_instance }}
     - enable: True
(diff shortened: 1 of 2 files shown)


netweaver/monitoring.sls
```diff
@@ -14,8 +14,13 @@ sap_host_exporter_exporter_service:
       - pkg: prometheus_sap_host_exporter_pkg
       - file: sap_host_exporter_configuration
 
-# for each node the instance number changes so we need to adapt configuration
+# the instance number of a node is unique and changes so we need to adapt configuration
```
arbulu89: As a comment sid,instance number pair is unique

netweaver/monitoring.sls
```diff
@@ -14,8 +14,13 @@ sap_host_exporter_exporter_service:
...
 {% for node in netweaver.nodes if host == node.host %}
+
+{% set instance = node.instance %}
```
arbulu89: Just to the same we do to get the instance in every place else:
```{% set instance = '{:0>2}'.format(node.instance) %}```

File: netweaver/monitoring.sls [hunk 3]
```diff
@@ -14,8 +14,13 @@ sap_host_exporter_exporter_service:
...
+{% set instance_name = node.sid~'_'~instance %}
+
+sap_host_exporter_configuration_{{ instance_name }}:
```
arbulu89: I guess you will need to remove the next line, right?

arbulu89 (changes requested):
@MalloZup Fix the changes please

netweaver/monitoring.sls
arbulu89: Actually the comment doesn't make sense, Rewrite too:
```
# the sid, instance number pair of a node is unique, so we need to adapt configuration
```

arbulu89 (changes requested):
@MalloZup Actually, you will need to do some adjustnments. For example, the DB instance is not SAP instance that starts any process, so there we don't need to start the exporter.
You will need to think about it. And in fact, this will create exporters collading the tcp port, as you are not changing. Many things to think about

netweaver/monitoring.sls
MalloZup: ```'{:0>2}'``` what is doing this by curiosity?

netweaver/monitoring.sls
MalloZup: thx

netweaver/monitoring.sls
```diff
@@ -1,26 +1,36 @@
...
-{% for node in netweaver.nodes if host == node.host %}
-sap_host_exporter_configuration:
+# we can have multiple exporters in the same host in that moment, we need to change the port used by the exporter
+sap_host_exporter_configuration_{{ daemon_instance }}:
   file.managed:
-    - name: /etc/sap_host_exporter.yaml
+    - name: /etc/sap_host_exporter{{ daemon_instance }}.yaml
     - contents: |
-         port: "9680"
+         port: "968{{ loop.index }}"
```
MalloZup: I think here is the only way to update the port. imho @arbulu89

MalloZup:
@arbulu89  I think now should be look fine. I will let this in wip until I develop the rest of the code on the exporter but the logic should fine now :football:

netweaver/monitoring.sls
arbulu89: @MalloZup I think `loop.index` stands for 1. If you want to start from 0 you need to change to `loop.index0`

netweaver/monitoring.sls
```diff
@@ -1,26 +1,36 @@
...
 
+# the sid, instance number pair of a node is unique, so we need to adapt configuration
+{% for node in netweaver.nodes if host == node.host %}
+
+{% set instance = node.instance %}
+{% set instance = '{:0>2}'.format(node.instance) %}
+{% set daemon_instance = '{}_{}'.format(node.sid, instance) %}
+
+# don't install exporter for db instance
+{%- if not node.sap_instance == "db" %}
```
arbulu89: I would put this in the `for` loop declaration, to avoid just this if check.
Something like
```
{% for node in netweaver.nodes if host == node.host and node.sap_instance != "db" %}
```

netweaver/monitoring.sls
```diff
@@ -1,26 +1,36 @@
...
+
```
arbulu89: Update the state name, otherwise we can have the unique id issue again:
```
sap_host_exporter_exporter_service_{{ daemon_instance }}:
```

netweaver/monitoring.sls
```diff
@@ -1,26 +1,36 @@
...
 
+
 prometheus_sap_host_exporter_pkg:
   pkg.installed:
     - name: prometheus-sap_host_exporter
 
+# the sid, instance number pair of a node is unique, so we need to adapt configuration
+{% for node in netweaver.nodes if host == node.host %}
+
+{% set instance = node.instance %}
```
arbulu89: Remove this first line, it's not used

netweaver/monitoring.sls
```diff
@@ -1,26 +1,36 @@
...
 sap_host_exporter_exporter_service:
   service.running:
-    - name: prometheus-sap_host_exporter
+    - name: prometheus-sap_host_exporter@{{ daemon_instance }}
     - enable: True
     - restart: True
     - require:
       - pkg: prometheus_sap_host_exporter_pkg
-      - file: sap_host_exporter_configuration
+      - file: sap_host_exporter_configuration_{{ daemon_instance }}
```
arbulu89, line 20: Besides the `require` entry , you need to add the `file` state in a `watch` state in salt. Have a look in:
https://docs.saltstack.com/en/latest/ref/states/all/salt.states.service.html

netweaver/monitoring.sls
```diff
@@ -1,26 +1,36 @@
...
 
-# for each node the instance number changes so we need to adapt configuration
-{% for node in netweaver.nodes if host == node.host %}
-sap_host_exporter_configuration:
+# we can have multiple exporters in the same host in that moment, we need to change the port used by the exporter
+sap_host_exporter_configuration_{{ daemon_instance }}:
   file.managed:
-    - name: /etc/sap_host_exporter.yaml
+    - name: /etc/sap_host_exporter{{ daemon_instance }}.yaml
```
arbulu89: As @stefanotorresi suggest online, I would put the config file in a folder, not just directly in `/etc`
Something like:
`/etc/sap_host_exporter/HA1-00.yaml` or `/usr/etc/sap_host_exporter/HA1-00.yaml`
For that, you will need to change the exporter spec file to create those folders

netweaver/monitoring.sls
MalloZup: done thx

netweaver/monitoring.sls
MalloZup: why this? In the past we agreed already this for the `ha_cluster_exporter`. If we are going to change it, we will need to unify also ha_cluster_exporter. I mean it is something we have previously agreed on that, having the conf file without folder. I think is ok, there are other config file in `etc` from different apps.

arbulu89: @MalloZup This is what I understood from @stefanotorresi when we talked on the chat. Anyway, if this is how `ha_cluster_exporter` go ahead, and put in `/etc`.
PD: `hanadb_exporter` was implemented before and we started using `/etc/hanadb_exporter` from the beginning. I don't really know where this `agreement` is coming from.

arbulu89: And in fact, one of the reasons of having a folder instead of a unique file is that in this exporter, you might have several config files `sap_host_exporter_prd_00.yaml`, `sap_host_exporter_prd_01.yaml`, `sap_host_exporter_prd_02.yaml`, so haveing them all around `/etc` folders looks a bit ugly. That's my opinion anyway.

MalloZup: ok make sense. We can create a folder via salt.

netweaver/monitoring.sls
```diff
@@ -5,22 +5,30 @@ prometheus_sap_host_exporter_pkg:
...
     - name: prometheus-sap_host_exporter
 
-sap_host_exporter_exporter_service:
+# the sid, instance number pair of a node is unique, so we need to adapt configuration
+{% for node in netweaver.nodes if host == node.host and node.sap_instance != "db" %}
+{% set instance = node.instance %}
+{% set instance = '{:0>2}'.format(node.instance) %}
+{% set daemon_instance = '{}_{}'.format(node.sid, instance) %}
+
+sap_host_exporter_exporter_service_{{ daemon_instance }}:
```
arbulu89: `exporter` word duplicated in the state name

netweaver/monitoring.sls
```diff
@@ -5,22 +5,30 @@ prometheus_sap_host_exporter_pkg:
   pkg.installed:
     - name: prometheus-sap_host_exporter
 
-sap_host_exporter_exporter_service:
+# the sid, instance number pair of a node is unique, so we need to adapt configuration
+{% for node in netweaver.nodes if host == node.host and node.sap_instance != "db" %}
+{% set instance = node.instance %}
```
arbulu89: Still this first line `{% set instance = node.instance %}` must be removed, it doesn't do anything

netweaver/monitoring.sls
```diff
@@ -5,22 +5,30 @@ prometheus_sap_host_exporter_pkg:
...
+sap_host_exporter_configuration_{{ daemon_instance }}:
   file.managed:
-    - name: /etc/sap_host_exporter.yaml
+    - name: /etc/sap_host_exporter{{ daemon_instance }}.yaml
     - contents: |
-         port: "9680"
+         port: "968{{ loop.index0 }}"
          address: "0.0.0.0"
          log-level: "info"
          sap-control-url: "http://127.0.0.1:5{{ node.instance }}13"
```
arbulu89: Don't use the `{{ node.instance }}` if the value is `0` it will create the port `5013` what is incorrect.
Use the `{{ instance }}` that was created before fixing this issue

sapnwbootstrap-formula.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Thu Mar 12 12:24:56 UTC 2020 - Dario Maiocchi <dmaiocchi@suse.com>
+-  Version 0.2.7
+   * Fix non-unique ID name error for monitoring.sls state
```
arbulu89, line 3: I would add 
```
* Manage properly multiple exporters execution in the same host
```

arbulu89 (changes requested):
@MalloZup Still some fixes are needed.

netweaver/monitoring.sls
MalloZup: done thx

MalloZup:
should be ok now. I will wait the PR  on the exporter get merged and test it with this

sapnwbootstrap-formula.changes
```diff
@@ -1,13 +1,18 @@
...
 Mon Mar 16 11:02:26 UTC 2020 - Xabier Arbulu <xarbulu@suse.com>
 
 - Version 0.2.7
   * Add option to set custom parameters in the installation
     configuration files
   * Make the installation retry times customizable
-
 -------------------------------------------------------------------
-Tue Mar 10 09:58:57 UTC 2020 - Dario Maiocchi <dmaiocchi@suse.com>
+ue Mar 10 09:58:57 UTC 2020 - Dario Maiocchi <dmaiocchi@suse.com>
```
arbulu89: You have remofe a `T` from this line

arbulu89 (approved):
@MalloZup Now it looks fine! Just fix that removed letter in the `changes` file

MalloZup:
thx

PR #41, opened 2020-03-18
Feature/sapcar support

smpsgill:
Adding the support for sapcar extraction. This is initial version of formula, I have not tested with the deployment and I will update it once I complete the tests along with corresponding salt-shaptools PR.

### DIFF
--- a/netweaver/extract_installer.sls
+++ b/netweaver/extract_installer.sls
@@ -1,12 +1,22 @@
 {%- from "netweaver/map.jinja" import netweaver with context -%}
 {% set host = grains['host'] %}
 {% for node in netweaver.nodes if node.host == host and (netweaver.sapcar_exe_file is defined and netweaver.swpm_sar_file and netweaver.swpm_extract_dir is defined) %}
+{% set instance = '{:0>2}'.format(node.instance) %}
+{% set name = '{}_{}'.format(node.sid, instance) %}
 
-extract_installer_file_{{ node.host }}:
-    netweaver.installer_file_extracted:
-    - sapcar_exe_file: {{ node.install.sapcar_exe_file }}
-    - swpm_sar_file: {{ node.install.swpm_sar_file }}
-    - swpm_extract_dir: {{ netweaver.swpm_extract_dir }}
-    - options: -manifest SIGNATURE.SMF
+
+validate_sapcar_permission_{{ host }}_{{ name }}:
+    file.managed:
+    - name: {{ netweaver.sapcar_exe_file }}
+    - mode: 644
+
+extract_installer_file_{{ host }}_{{ name }}:
+    sapcar.extracted:
+    - name: {{ netweaver.swpm_sar_file }}
+    - sapcar_exe: {{ netweaver.sapcar_exe_file }}
+    - output_dir: {{ netweaver.swpm_extract_dir }}
+    - options: "-manifest SIGNATURE.SMF"
+    - require:
+      - validate_sapcar_permission_{{ host }}_{{ name }}
 
 {% endfor %}
--- a/netweaver/extract_installer.sls
+++ b/netweaver/extract_installer.sls
@@ -1,22 +1,13 @@
 {%- from "netweaver/map.jinja" import netweaver with context -%}
 {% set host = grains['host'] %}
-{% for node in netweaver.nodes if node.host == host and (netweaver.sapcar_exe_file is defined and netweaver.swpm_sar_file and netweaver.swpm_extract_dir is defined) %}
-{% set instance = '{:0>2}'.format(node.instance) %}
-{% set name = '{}_{}'.format(node.sid, instance) %}
 
+{% if netweaver.sapcar_exe_file is defined and netweaver.swpm_sar_file is defined %}
 
-validate_sapcar_permission_{{ host }}_{{ name }}:
-    file.managed:
-    - name: {{ netweaver.sapcar_exe_file }}
-    - mode: 644
-
-extract_installer_file_{{ host }}_{{ name }}:
+extract_installer_file:
     sapcar.extracted:
     - name: {{ netweaver.swpm_sar_file }}
     - sapcar_exe: {{ netweaver.sapcar_exe_file }}
     - output_dir: {{ netweaver.swpm_extract_dir }}
     - options: "-manifest SIGNATURE.SMF"
-    - require:
-      - validate_sapcar_permission_{{ host }}_{{ name }}
 
-{% endfor %}
+{% endif %}
--- a/netweaver/extract_installer.sls
+++ b/netweaver/extract_installer.sls
@@ -1,5 +1,4 @@
 {%- from "netweaver/map.jinja" import netweaver with context -%}
-{% set host = grains['host'] %}
 
 {% if netweaver.sapcar_exe_file is defined and netweaver.swpm_sar_file is defined %}
 
--- a/netweaver/extract_installer.sls
+++ b/netweaver/extract_installer.sls
@@ -1,0 +1,12 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+{% for node in netweaver.nodes if node.host == host and (netweaver.sapcar_exe_file is defined and netweaver.swpm_sar_file and netweaver.swpm_extract_dir is defined) %}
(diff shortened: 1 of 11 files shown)


arbulu89 (changes requested):
Hi @Simranpal ,
I don't really like this approach. I wouldn't have something like `netweaver.installer_file_extracted`.
Why do we need something specific for netweaver? What we need is just a method that extracts sar files. Nothing `netweaver` nor `hana` specific.

I would create a new salt state for that. Something like:
```
extract_installer_file_{{ sid+instance+host }}: # don't use host, as 1 node can have multiple sap
    sapcar.file_extracted:
        - name: {{ sapcar_executable_path }}
        - files: # It would be nice if we give the option to extract multiple files, anyway, in the 1st iteraction we can just support 1 file extraction
            - my_first_sarfile
            - my_second_sarfile
        - output_folder: /netweaver_inst_media
        - options: "-manifest SIGNATURE.SMF"
```

So the idea is that the method is reusable for any context where we need to extract sar files (HANA and NETWEAVER by now). 

2. As most of the times all of the nodes will use the same files, I would like to have the option to set a global extraction data outside the nodes (we already do this for other things). So this way, if we have global data, it's used in all the nodes, and if we have this data in the node itself, it will have preference. The idea is to have the option to have the option to create a smaller pillar file.
Anyway, the current approach is good enough by now, set the same data for all the nodes.

netweaver/extract_installer.sls
```diff
@@ -0,0 +1,12 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+{% for node in netweaver.nodes if node.host == host and (netweaver.sapcar_exe_file is defined and netweaver.swpm_sar_file and netweaver.swpm_extract_dir is defined) %}
+
+extract_installer_file_{{ node.host }}:
+    netweaver.installer_file_extracted:
+    - sapcar_exe_file: {{ node.install.sapcar_exe_file }}
+    - swpm_sar_file: {{ node.install.swpm_sar_file }}
+    - swpm_extract_dir: {{ netweaver.swpm_extract_dir }}
```
arbulu89: Make sure before running this method that the `outpur_dir` exists. You will need to add a new `file.directory` state before I guess

arbulu89: And besides that, we need to make sure that the SAPCAR has executables rights, `755` file permissions work, but I don't know if there is some recommendation for that.
So, we will need to make sure that the file is executable before running the new methods.

arbulu89: Other things. I have find out that the SIGNATURE.SMF is created in the folder where SAPCAR is executed (or -R folder), but I would say thay we will need to copy the created file to the folder where the installer is located, so we will need to create an additional state step to copy the file

smpsgill: Hi @arbulu89 the `outpur_dir` folder gets created by sapcar if it doesnt exist & overridden if exists.
I have added step to check the sapcar permissions.

smpsgill: Regarding the `SIGNATURE.SMF` incase of NW SWPM installer, this is extracted correctly to the location where `sapinst` is present.
We would need to copy that in case of HANA installer, maybe in the HANA formula that can be done.

netweaver/extract_installer.sls
```diff
@@ -0,0 +1,22 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+{% for node in netweaver.nodes if node.host == host and (netweaver.sapcar_exe_file is defined and netweaver.swpm_sar_file and netweaver.swpm_extract_dir is defined) %}
+{% set instance = '{:0>2}'.format(node.instance) %}
+{% set name = '{}_{}'.format(node.sid, instance) %}
+
```
arbulu89, line 2: Why so many new lines?

netweaver/extract_installer.sls
```diff
@@ -0,0 +1,22 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+{% for node in netweaver.nodes if node.host == host and (netweaver.sapcar_exe_file is defined and netweaver.swpm_sar_file and netweaver.swpm_extract_dir is defined) %}
```
arbulu89: We don't need this loop, right? It's quite useless. I would just remove it. The code will do the same without it. In fact, with the loop, the states will be executed needlessly multiple times, potentially creating issues.

netweaver/extract_installer.sls
```diff
@@ -0,0 +1,22 @@
...
+{% set instance = '{:0>2}'.format(node.instance) %}
+{% set name = '{}_{}'.format(node.sid, instance) %}
+
+
+validate_sapcar_permission_{{ host }}_{{ name }}:
```
arbulu89: I don't know if we can do this. This file might be in NFS shared folder where users don't have write permissions.

arbulu89 (changes requested):
@Simranpal As first thoughts.

I would set a default value to `swpm_extract_dir`. This would make things easier.
2nd, I will research about jinja/salt macros to see if we can have these duplicated piece of new code in a macro/method/something like that. It would make the code neater

netweaver/extract_installer.sls
smpsgill: For the SAPCAR permission, there is no recommendation as such that I could find in the docs/notes. Will it not be user responsibility to provide latest version(I will update that in the docs) and have correct permissions to executable file?
We can verify if the permissions are correct(via salt module fn) before executing the methods. Or copy it to a temp location from where we can execute it with the correct rights. What do you think about this @arbulu89 @diegoakechi

netweaver/extract_installer.sls
smpsgill: Thanks, there is extra line from previous removed line. I corrected this.

netweaver/extract_installer.sls
smpsgill: Yes, this state will execute on each host and we have the global extraction data which is common to all hosts, so its fine to remove it.

However, removing the loop, also takes away ability to set the state name based of `sid` and `instance` from node pillar entries. And we do not have those entries in the grains. Should I just use host in the state name, only issue will be when 1 host having multiple instances.

arbulu89: Why do you need any `sid`, `instance` or `host` ? The reason why this items are added to the state names is just to avoid duplicated state names, only because you loop in a machine. If you don't need to loop (because the action must be performed directly in all nodes), you can have just the same name for the state.

Just put `extract_installer_file` and that's it. This works totally fine. We don't need to add not needed extra complexity.

netweaver/extract_installer.sls
arbulu89: mmm, we can think about it in the future, maybe by now checking the permissions is correct.
I don't know @Simranpal , this is a trial an error thing.
Just run the test in our libvirt, azure and aws deployments and see what happens.
I cannot really tell you more.
This is how I did in many other occasions. Try and see the result, and implement according that.

diegoakechi: @Simranpal @arbulu89 the OS already checks the permission and if the error is readable enough I would simple delegate this to the OS and raise the execution error if it occurs to avoid extra checks (since it will be checked by the OS anyway). Documenting that the sapcar binary provided needs to have execution permissions seem enough for me.

As you have already implemented the check, I think we can keep it, but make sure to test if it will not fail in case the file has different permission, like 755 that would still be executable but AFAIK salt will try to change the permissions. Imagine that the user can point a NFS share where sapcar is mounted with read-only access, so in that situation, I think the validation would fail the execution.

smpsgill:
TODO: Update spec,change files after testing with the NW deployment

netweaver/extract_installer.sls
```diff
@@ -0,0 +1,13 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
```
arbulu89: I think we can remove this `host` variable line

arbulu89 (approved):
@Simranpal It looks fine. Just test with and without SAR file, and if it works, update the spec/changes file and merge

smpsgill:
After testing with Netweaver deployments, this changes worked fine

PR #42, opened 2020-03-20
Feature/install pydbapi

arbulu89:
Until now, we were using PyHDB to connect to the HANA database during the installation. PyHDB is not an official package. SAP provides the official python SAP Client package, that is needed to install the DB, PAS and AAS instances.
Taking advantage that this software is mandatory, this change extracts and install the python api.

### DIFF
--- a/netweaver/setup/install_pydbapi.sls
+++ b/netweaver/setup/install_pydbapi.sls
@@ -1,0 +1,46 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+{% set host = grains['host'] %}
+
+{% for node in netweaver.nodes if node.host == host and node.sap_instance in ['db', 'pas', 'aas'] %}
+{% if loop.first %}
+{% set pydbapi_output_dir = '/tmp/pydbapi' %}
+nw_install_python_pip:
+  pkg.installed:
+    {% if grains['pythonversion'][0] == 2 %}
+    - name: python-pip
+    {% else %}
+    - name: python3-pip
+    {% endif %}
+    - retry:
+        attempts: 3
+        interval: 15
+    - resolve_capabilities: true
+
+nw_extract_pydbapi_client:
+  hana.pydbapi_extracted:
+    - name: PYDBAPI.TGZ
+    - software_folders: {{ netweaver.additional_dvds }}
+    - output_dir: {{ pydbapi_output_dir }}
+    - hana_version: '20'
+    - force: true
+
+# pip.installed fails as it cannot manage propler file names with regular expressions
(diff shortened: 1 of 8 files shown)


netweaver/setup/install_pydbapi.sls
```diff
@@ -0,0 +1,53 @@
...
+        attempts: 3
+        interval: 15
+    - resolve_capabilities: true
+
+nw_extract_pydbapi_client:
+  hana.pydbapi_extracted:
+    - name: PYDBAPI.TGZ
+    - software_folders: {{ netweaver.additional_dvds }}
+    - output_dir: {{ pydbapi_output_dir }}
+    - hana_version: '20'
```
diegoakechi, line 24: @arbulu89 why do you block explicitly to HANA 2 here? I think for HANA 1 the process was exactly the same...

netweaver/setup/install_pydbapi.sls
```diff
@@ -0,0 +1,53 @@
...
+    - force: true
+
+# pip.installed fails as it cannot manage propler file names with regular expressions
+# TODO: Improve this to use pip.installed somehow
+nw_install_pydbapi_client:
```
diegoakechi, line 29: To simplify this, maybe you could search the exact file name first and then call the pip state.

diegoakechi (changes requested):
Small nit pick

netweaver/setup/install_pydbapi.sls
arbulu89: The `20` is used during the automatic finding of the folder. The patter of the HANA client is different for HANA 1 and HANA 2. We need to somehow specify this. Otherwise, but happens, if somehow you provide HANA 1 and HANA 2 software folders? It would just return the first match.
The discussion should be if this could be paremetrized. And yes, it could be. I could just set other pillar variable putting `20` as default

netweaver/setup/install_pydbapi.sls
arbulu89: Is this simpler or more complicated? hehe
Remember that we cannot store variables from salt execution modules. So, I would need to:
- Create a new execution module to do so
- Create some ugly bash script line to find the name (I don't know if this would work)

netweaver/setup/install_pydbapi.sls
diegoakechi: I think it is fine for now.

netweaver/setup/install_pydbapi.sls
diegoakechi: I didn't remember because I didn't know 😆

diegoakechi (approved):

PR #43, opened 2020-03-31
Feature: exporter resource

stefanotorresi:
This PR implements Pacemaker resources for the sap_host_exporter instances to go along with the SAP instances they monitor.

### DIFF
--- a/netweaver/monitoring.sls
+++ b/netweaver/monitoring.sls
@@ -15,12 +15,8 @@ sap_host_exporter_configuration_{{ exporter_instance }}:
   file.managed:
     - name: /etc/sap_host_exporter/{{ exporter_instance }}.yaml
     - contents: |
-         port: "9680"
          address: {{ node.virtual_host }}
-         log-level: "info"
-         sap-control-url: "http://127.0.0.1:5{{ sap_instance_nr }}13"
-         sap-control-user: "{{ node.sid.lower() }}adm"
-         sap-control-password: "{{ netweaver.sid_adm_password|default(netweaver.master_password) }}"
+         sap-control-uds: /tmp/.sapstream5{{ sap_instance_nr }}13
     - require:
       - pkg: prometheus_sap_host_exporter_pkg
 
--- a/netweaver/monitoring.sls
+++ b/netweaver/monitoring.sls
@@ -6,33 +6,34 @@ prometheus_sap_host_exporter_pkg:
     - name: prometheus-sap_host_exporter
 
 # the sid, instance number pair of a node is unique, so we need to adapt configuration
-{% for node in netweaver.nodes if host == node.host and node.sap_instance != "db" %}
-{% set instance = '{:0>2}'.format(node.instance) %}
-{% set daemon_instance = '{}_{}'.format(node.sid, instance) %}
+{% for node in netweaver.nodes if node.sap_instance != "db" %}
+{% set sap_instance_nr = '{:0>2}'.format(node.instance) %}
+{% set exporter_instance = '{}_{}{}'.format(node.sid, node.sap_instance.upper(), sap_instance_nr) %}
 
-sap_host_exporter_service_{{ daemon_instance }}:
-  service.running:
-    - name: prometheus-sap_host_exporter@{{ daemon_instance }}
-    - enable: True
-    - restart: True
-    - require:
-      - pkg: prometheus_sap_host_exporter_pkg
-      - file: sap_host_exporter_configuration_{{ daemon_instance }}
-    - watch:
-      - file: sap_host_exporter_configuration_{{ daemon_instance }}
-
-
-# we can have multiple exporters in the same host in that moment, we need to change the port used by the exporter
-sap_host_exporter_configuration_{{ daemon_instance }}:
+# we bind each exporter instance to a SAP instance virtual host
+sap_host_exporter_configuration_{{ exporter_instance }}:
   file.managed:
-    - name: /etc/sap_host_exporter/{{ daemon_instance }}.yaml
+    - name: /etc/sap_host_exporter/{{ exporter_instance }}.yaml
     - contents: |
-         port: "968{{ loop.index0 }}"
-         address: "0.0.0.0"
+         port: "9680"
+         address: {{ node.virtual_host }}
          log-level: "info"
-         sap-control-url: "http://127.0.0.1:5{{ instance }}13"
+         sap-control-url: "http://127.0.0.1:5{{ sap_instance_nr }}13"
          sap-control-user: "{{ node.sid.lower() }}adm"
          sap-control-password: "{{ netweaver.sid_adm_password|default(netweaver.master_password) }}"
     - require:
       - pkg: prometheus_sap_host_exporter_pkg
+
+{% if host == node.host %}
+sap_host_exporter_service_{{ exporter_instance }}:
+  service.running:
+    - name: prometheus-sap_host_exporter@{{ exporter_instance }}
+    - enable: {{ not netweaver.ha_enabled }}
+    - restart: True
+    - require:
+      - pkg: prometheus_sap_host_exporter_pkg
+      - file: sap_host_exporter_configuration_{{ exporter_instance }}
+    - watch:
+      - file: sap_host_exporter_configuration_{{ exporter_instance }}
+{% endif %}
 {% endfor %}
--- a/templates/cluster_resources.j2
+++ b/templates/cluster_resources.j2
@@ -134,8 +134,12 @@ group grp_{{ sid }}_ASCS{{ ascs_instance }} \
   rsc_ip_{{ sid }}_ASCS{{ ascs_instance }} \
   rsc_fs_{{ sid }}_ASCS{{ ascs_instance }} \
   rsc_sap_{{ sid }}_ASCS{{ ascs_instance }} \
-  {% if cloud_provider == "microsoft-azure" -%} rsc_socat_{{ sid }}_ASCS \ {%- endif %}
-  {% if monitoring_enabled -%} rsc_exporter_{{ sid }}_ASCS{{ ascs_instance }} \ {%- endif %}
+  {%- if cloud_provider == "microsoft-azure" %}
+  rsc_socat_{{ sid }}_ASCS \
+  {%- endif %}
+  {%- if monitoring_enabled %}
+  rsc_exporter_{{ sid }}_ASCS{{ ascs_instance }} \
+  {%- endif %}
   meta resource-stickiness=3000
 
 primitive rsc_fs_{{ sid }}_ERS{{ ers_instance }} Filesystem \
@@ -159,8 +163,12 @@ group grp_{{ sid }}_ERS{{ ers_instance }} \
   rsc_ip_{{ sid }}_ERS{{ ers_instance }} \
   rsc_fs_{{ sid }}_ERS{{ ers_instance }} \
   rsc_sap_{{ sid }}_ERS{{ ers_instance }} \
-  {% if cloud_provider == "microsoft-azure" -%} rsc_socat_{{ sid }}_ERS \ {%- endif %}
-  {% if monitoring_enabled -%} rsc_exporter_{{ sid }}_ERS{{ ers_instance }} \ {%- endif %}
+  {%- if cloud_provider == "microsoft-azure" %}
+  rsc_socat_{{ sid }}_ERS \
+  {%- endif %}
+  {%- if monitoring_enabled -%}
+  rsc_exporter_{{ sid }}_ERS{{ ers_instance }} \
+  {%- endif %}
   meta resource-stickiness=3000
 
 colocation col_sap_{{ sid }}_no_both -5000: grp_{{ sid }}_ERS{{ ers_instance }} grp_{{ sid }}_ASCS{{ ascs_instance }}
--- a/templates/cluster_resources.j2
+++ b/templates/cluster_resources.j2
@@ -11,6 +11,7 @@
 {%- set ers_ip_address = data.ers_ip_address %}
 {%- set ascs_virtual_host = data.ascs_virtual_host %}
 {%- set ers_virtual_host = data.ers_virtual_host %}
+{%- set monitoring_enabled = data.monitoring_enabled|default(False) %}
 
 # Platform dependant (stonith, virtual ip address, cib options, etc) resource
 
@@ -93,7 +94,7 @@ primitive rsc_ip_{{ sid }}_ERS{{ ers_instance }} IPaddr2 \
 
 {%- endif %}
 
-{%- if data.monitoring_enabled %}
+{%- if monitoring_enabled %}
 
 primitive rsc_exporter_{{ sid }}_ASCS{{ ascs_instance }} systemd:prometheus-sap_host_exporter@{{ sid }}_ASCS{{ ascs_instance }} \
     op start interval=0 timeout=100 \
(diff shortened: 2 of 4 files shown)


templates/cluster_resources.j2
```diff
@@ -77,6 +77,22 @@ primitive rsc_ip_{{ sid }}_ERS{{ ers_instance }} IPaddr2 \
 
 {%- endif %}
 
+{%- if data.monitoring_enabled %}
+
+primitive rsc_exporter_{{ sid }}_ASCS{{ ascs_instance }} systemd:prometheus-sap_host_exporter@{{ sid }}_ASCS{{ ascs_instance }} \
+    op start interval=0 timeout=60 \
```
arbulu89: @stefanotorresi As a comment, `crmsh` complains that the `start/stop` timeouts are smaller than the recommended 100 value. We should change it in the ASCS and ERS exporter

```
WARNING: rsc_exporter_HA1_ASCS00: specified timeout 60 for start is smaller than the advised 100
WARNING: rsc_exporter_HA1_ASCS00: specified timeout 60 for stop is smaller than the advised 100
WARNING: rsc_exporter_HA1_ERS10: specified timeout 60 for start is smaller than the advised 100
WARNING: rsc_exporter_HA1_ERS10: specified timeout 60 for stop is smaller than the advised 100
```

netweaver/monitoring.sls
```diff
@@ -6,33 +6,34 @@ prometheus_sap_host_exporter_pkg:
...
+# we bind each exporter instance to a SAP instance virtual host
+sap_host_exporter_configuration_{{ exporter_instance }}:
   file.managed:
-    - name: /etc/sap_host_exporter/{{ daemon_instance }}.yaml
+    - name: /etc/sap_host_exporter/{{ exporter_instance }}.yaml
     - contents: |
-         port: "968{{ loop.index0 }}"
-         address: "0.0.0.0"
+         port: "9680"
+         address: {{ node.virtual_host }}
```
arbulu89, line 18: Small consideration with this. Attaching the virtual host to the exporter will make the endpoint accessible only for this network, so only in this network. We are not going to be able to access the metrics from our computers.
For real usage is fine, but for development is annoying.
We cannot really do anything about this, but to keep in mind the restriction.

stefanotorresi: well, there are a few ways around this:
- you can log in the node and browse the metrics locally with curl
- you can use the Prometheus dashboard

it's a small tradeoff to avoid some quirky port allocation logic within the salt state.

templates/cluster_resources.j2
```diff
@@ -98,7 +114,9 @@ primitive rsc_sap_{{ sid }}_ASCS{{ ascs_instance }} SAPInstance \
      migration-threshold=1 priority=10
 
 group grp_{{ sid }}_ASCS{{ ascs_instance }} \
-  rsc_ip_{{ sid }}_ASCS{{ ascs_instance }} rsc_fs_{{ sid }}_ASCS{{ ascs_instance }} rsc_sap_{{ sid }}_ASCS{{ ascs_instance }} {% if cloud_provider == "microsoft-azure" %} nc_{{ sid }}_ASCS {% endif %} \
+  rsc_ip_{{ sid }}_ASCS{{ ascs_instance }} rsc_fs_{{ sid }}_ASCS{{ ascs_instance }} rsc_sap_{{ sid }}_ASCS{{ ascs_instance }}
+  {%- if cloud_provider == "microsoft-azure" %} nc_{{ sid }}_ASCS {% endif -%}
+  {%- if data.monitoring_enabled %} rsc_exporter_{{ sid }}_ASCS{{ ascs_instance }} {% endif -%}
```
arbulu89: I increased the `resource-stickiness` to `4000` in my tests and works fine. Need to really understand what we need to put here

stefanotorresi: I don't quite understand, what makes 4000 better than the old value 3000?

templates/cluster_resources.j2
```diff
@@ -77,6 +77,22 @@ primitive rsc_ip_{{ sid }}_ERS{{ ers_instance }} IPaddr2 \
 
 {%- endif %}
 
+{%- if data.monitoring_enabled %}
```
arbulu89: I would set `false` by default to this option, so we don't force the user to put it. In fact, having an exporter is optional

templates/cluster_resources.j2
arbulu89: To be honest, no idea... I remember some days ago that @diegoakechi told us something about this. As far as the migration is done, I don't really see any difference.

PD: I have tested with 3000 and the behaviour looks the same

Info:
- https://clusterlabs.org/pacemaker/doc/en-US/Pacemaker/1.1/html/Pacemaker_Explained/s-resource-options.html
- https://clusterlabs.org/pacemaker/doc/en-US/Pacemaker/1.1/html/Clusters_from_Scratch/_prevent_resources_from_moving_after_recovery.html

templates/cluster_resources.j2
stefanotorresi: Hmm, I don't follow. There is no default here (which means the expression will evaluate to false unless the pillar is set to non-false value).

The default is set in SUSE/ha-sap-terraform-deployments#406.

arbulu89: Are you sure about that? I'm not totally. I think you would have an error like this if the `monitoring_enabled` is not in the `data` dictionary
```
Unable to manage file: Jinja variable 'salt.utils.odict.OrderedDict object' has no attribute 'monitoring_enabled'
```

What I mean is that I would put this as `{%- if data.monitoring_enabled|default(false) %}`, to make this parameter optional. Otherwise, we are forcing to have `true` or `false` (or something at least) in this entry. At this moment the parameter is not optional anymore.

PD: And the default value in the terraform project is not enough. Think about this formulas as an independent being, not as something that will work always with that project.

stefanotorresi: Ohh, I see what you mean, now. So, the `|ḍefault` filter also acts as a null coalesce operation? Yeah, I can add that no-problem.

templates/cluster_resources.j2
diegoakechi: @stefanotorresi @arbulu89 It should be ok to keep 3000, as the group receives 3000 not the resource. The problem I had was on the yast module, since it configures the stickiness to 1000, so every new resource sum more 1000 on the stickiness calculation, but it got improved on the current best practices.

stefanotorresi: so should I leave it to 3000 then?

diegoakechi: On my POV, yes, but it is more important to test it and see if works. Basically the resources on the group needs to migrate together with the ASCS instance in case of failover. If it works, to 3000 is fine.

stefanotorresi:
@arbulu89 I was wondering: since we now have UDS support in `sap_host_exporter`, should we use the socket in the configuration, rather than using TCP/IP w/ authentication? I think it would simplify things and it should also be faster.

templates/cluster_resources.j2
stefanotorresi: okay, thanks for your input Diego! It appears to work quite well, I will test this further.

templates/cluster_resources.j2
```diff
@@ -119,7 +138,10 @@ primitive rsc_sap_{{ sid }}_ERS{{ ers_instance }} SAPInstance \
   meta priority=1000
 
 group grp_{{ sid }}_ERS{{ ers_instance }} \
-  rsc_ip_{{ sid }}_ERS{{ ers_instance }} rsc_fs_{{ sid }}_ERS{{ ers_instance }} rsc_sap_{{ sid }}_ERS{{ ers_instance }} {% if cloud_provider == "microsoft-azure" %} nc_{{ sid }}_ERS {% endif %}
+  rsc_ip_{{ sid }}_ERS{{ ers_instance }} rsc_fs_{{ sid }}_ERS{{ ers_instance }} rsc_sap_{{ sid }}_ERS{{ ers_instance }}
```
MalloZup: is it possible to add a comment on this ( i mean all the template file looks human unreadable and ulgy).. :man_shrugging:

templates/cluster_resources.j2
```diff
@@ -77,6 +78,22 @@ primitive rsc_ip_{{ sid }}_ERS{{ ers_instance }} IPaddr2 \
 
 {%- endif %}
 
+{%- if monitoring_enabled %}
```
MalloZup, line 97: a comment on the highelve of both resource might appreaciate

MalloZup:
@stefanotorresi we should not hardcode `uds` or `http` . We should offer both here, a machinery mechanism. Historically it was `http` only here because it existed only that

We need to offer both, and then it is the `deployment` project which will set the needed parameters wanted by user. Here is only machinery, so we should implement both in the formula.

MalloZup:
for example, take look here from pillar.example (https://github.com/SUSE/sapnwbootstrap-formula/blob/master/pillar.example#L80)
```
  # optional: enable sap_host_exporter (disabled by default)
  # the exporter will be installed and configured in all the nodes if it is enabled
  sap_host_exporter: 
    enabled: false
   ```

we need  something like 
```
  # optional: enable sap_host_exporter (disabled by default)
  # the exporter will be installed and configured in all the nodes if it is enabled
  sap_host_exporter: 
    enabled: false
    mode: http # or uds
```

**Comment by stefanotorresi** [comment]
I'm surprised by this take, you usually fancy being a minimalist! Why would we need something like this? Do we have an actual use case to give users this choice? There are a lot of other options that we don't expose, and this looks like a very internal detail that users should not need to care about.

File: templates/cluster_resources.j2 (hunk 6 above)
Comment (stefanotorresi) [reply templates/cluster_resources.j2:97]: what's not clear about this if condition?

File: templates/cluster_resources.j2 (hunk 5 above)
Comment (stefanotorresi) [reply templates/cluster_resources.j2]: I'm not sure what are you asking here. What should I comment? I don't even know what's the comment syntax on a crmsh config file. :D

**Comment by MalloZup** [comment]
the formula by design hould be configurable and expose most all functionality.
this can be than selected on the ha-deployment project as end-user

 In our context, exposing the mode is important.
So in my mind, the `mode` uds or http is really a core one which we need to expose.
We can add some defaults, if not specified but I consider that the endsolution like the `ha-sap-deployment` project should have the possiblity to change the pillar to `uds` or `http`, so like a sysadmin can choose without to handle the exporter configuration which functionality he prefer to use.

Otherwise, from highlevel pov the functionality doesn't exists. I wouldn't pretend that the sysadmin install and configure the exporter by hands. 
Maybe there is a target for sure, but that is why we need to offer both imho. Otherwise hardcoding one doesn't provide values in this context

File: templates/cluster_resources.j2 (hunk 6 above)
Comment (MalloZup) [reply templates/cluster_resources.j2:97]: not the if condition but the block.. anyways if there is no comment possible on that language is ok. I mean just from external pov, is really un-readable the thing , if you don't have code it you don't know what it does :grin:

File: templates/cluster_resources.j2 (hunk 5 above)
Comment (MalloZup) [reply templates/cluster_resources.j2]: ok if ther is no comment possible is ok

File: templates/cluster_resources.j2 (hunk 6 above)
Comment (stefanotorresi) [reply templates/cluster_resources.j2:97]: oh come on, it's not that bad, you just have to know a little about how `crmsh` works! I didn't know how to do any of this before this PR.
What the block does is it adds two new systemd resources, one for each exporter that is co-located with a HA SAP resource.

**Comment by stefanotorresi** [comment]
I wouldn't go as far as making this configurable in the deployment projects, really. We already have too many variables there.

I might agree to make the formula, and just the formula, more configurable via pillars, but I'd like to know why do we want to make this particular setting configurable in the first place. The formula installs the exporter in an opinionated way: along with the SAP instances.
The only reason to not use UDS and use TCP/IP instead is when the exporter is not colocated and runs in a different host, which cannot happen when using this formula, so what is the point of introducing an abstraction over the "mode" of operation of the exporter, if one of the options is basically useless in this context?

File: templates/cluster_resources.j2 (hunk 6 above)
Comment (MalloZup) [reply templates/cluster_resources.j2:97]: I mean this whole commands are ugly, but I don't feel it mandatory. It is a nitpick but  ok. I'm alwawys for putting rendundant comment which can safe the live for not insiders but ok, I don't expect this

**Comment by MalloZup** [comment]
why we don't remove the `http` support if we just use sockets, if it is useless? (taking the contra argument)

I would agree on this, if we remove http, hardcoding. 
 But if we offer both mode, we need a parameter on the formula.

We can set a default parameter on the formula but we need a mechanism to choose one of both.

Anyways, the formula are already configurable on the deployment project. Is just a matter of changing the pillars

**Comment by stefanotorresi** [comment]
Mate, we're talking about the formula here, not the exporter. Let's not conflate the two things together!
The exporter has its own set of features. The TCP/IP support is already there, there is no reason to remove it, especially because it's useful in non-HA deployments and also during development.

This formula has another responsibility: it install and configures the exporter in a HA deployment. In this context, the TCP/IP feature is not needed.

This formula is not the only way to install the exporter, so it doesn't make sense to remove a feature of another project because it's not used by one consumer of the project!

**Comment by diegoakechi** [comment]
@MalloZup It is a matter of trade-off. By default, Webservices are supposed to use HTTP, so this is what we assume as default here. You can be sure that NW or S/4 HANA also have many other options that we don't expose for simplicity here. UDS is a technical option of the exporter, but not being the default/standard for Webservices, it might represent a really small portion of the user base, so one extra option can make all the standard users to "pay" the price of deal with an extra option in favor of a very small minority that has this specific need. So for now, let's not overcomplicate the formulas and leave this option out. If this becomes a real use-case/customer request, it has a fast implementation anyway.

**Comment by MalloZup** [comment]
@diegoakechi  we are making here `UDS` as standard  right now and removing the https as default as it was before.

SO this imho is not a good step. A part of that, I still think that we should be implementing the low-hang fruit of the mode. UDS is not standard to set as default.

but personally ok, I don't think we are over-complicated the formula. 
The formula itself are an inventory with paramaters. 

So imho having default UDS might be problematic for all environment.

**Review by MalloZup — APPROVED** [review]
ok for moment we can see later how it goes .thx

File: netweaver/monitoring.sls [hunk 7]
```diff
@@ -6,33 +6,30 @@ prometheus_sap_host_exporter_pkg:
...
-    - name: prometheus-sap_host_exporter@{{ daemon_instance }}
-    - enable: True
-    - restart: True
+# we bind each exporter instance to a SAP instance virtual host
+sap_host_exporter_configuration_{{ exporter_instance }}:
+  file.managed:
+    - name: /etc/sap_host_exporter/{{ exporter_instance }}.yaml
+    - contents: |
+         address: {{ node.virtual_host }}
+         sap-control-uds: /tmp/.sapstream5{{ sap_instance_nr }}13
```
arbulu89, line 19: Is this `/tmp` file usage reliable? I mean, the content in this folder is removed periodically. Can we trust that this solution will work long term?

arbulu89 (changes requested):
The code looks fine. If it's tested and works, just update the `spec` and `changes` files to upgrade the version.

netweaver/monitoring.sls
stefanotorresi: This is how SAP documents UDS usage. I think the socket files are created by the SAP Start Service.

https://wiki.scn.sap.com/wiki/display/SI/How+to+find+the+client+of+a+UDS+connection

stefanotorresi:
@arbulu89 I now fully tested this and it works like a charm. Should be good to merge.

arbulu89 (approved):
Great job @stefanotorresi 
Thanks. We can merge now.

PR #46, opened 2020-04-22
Make it optional to install netweaver required packages

smpsgill:
The netweaver formula should also have the feature of netweaver required packages to be installed optionally.

As discussed in https://github.com/SUSE/ha-sap-terraform-deployments/pull/444 this option is also required for `qa_mode` which does not require these packages to be installed always.
I have set the default option value to `true` for installing packages.

### DIFF
--- a/sapnwbootstrap-formula.changes
+++ b/sapnwbootstrap-formula.changes
@@ -2,8 +2,7 @@
 Wed Apr 22 19:17:03 UTC 2020 - Simranpal Singh <simranpal.singh@suse.com>
 
 - Version 0.4.1
-  * Make it optional to install netweaver required packages 
-    in order to support qa mode
+  * Make it optional to install netweaver required packages
 
 -------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


arbulu89 (approved):
@Simranpal It looks fine. Please, test if properly before merging.
I don't know if having variable `netweaver` with the same name like in `netweaver.setup.packages` might cause issues.

sapnwbootstrap-formula.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Apr 22 19:17:03 UTC 2020 - Simranpal Singh <simranpal.singh@suse.com>
+
+- Version 0.4.1
+  * Make it optional to install netweaver required packages 
+    in order to support qa mode
```
arbulu89: @Simranpal I didn't see this before. Remove any reference to `qa_mode` please. This is nothing that this formula should know. The 1st sentence is good enough

smpsgill:
Tested with netweaver deployment and worked as expected

PR #47, opened 2020-05-05
Refactor monitoring settings

stefanotorresi:
Part of a refactoring spread among all the formulas dealing with monitoring settings.

See also SUSE/ha-sap-terraform-deployments#465.

### DIFF
--- a/netweaver/defaults.yaml
+++ b/netweaver/defaults.yaml
@@ -13,6 +13,4 @@ netweaver:
       directory: /var/lib/swap
       # Swap size in MB
       size: 2048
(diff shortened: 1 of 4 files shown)


arbulu89 (approved):
@stefanotorresi Thanks. I hope we don't need to add any kind of particular customization for the exporter.
Please upgrade package version and changes file before merging

arbulu89:
@stefanotorresi I think we need to change the template to be aligned with the other formulas:
https://github.com/SUSE/sapnwbootstrap-formula/blob/master/templates/cluster_resources.j2#L14

PR #48, opened 2020-05-12
Fix shared disk usage in non HA environments

arbulu89:
`shared_disk` usage is not needed in non HA environments (there is not any need to mount the ASCS data in a shared disk).

This changes the need to have this parameter yes or yes, now it can be ommited for non HA environments.

I have added some more comments in the pillar file to show how it works

### DIFF
--- a/netweaver/setup/mount.sls
+++ b/netweaver/setup/mount.sls
@@ -7,6 +7,13 @@
 
 {% if node.sap_instance.lower() in ['ascs', 'ers'] %}
 
+create_folder_{{ node.sap_instance.lower() }}_{{ instance_name }}:
+  file.directory:
+    - name: /usr/sap/{{ node.sid.upper() }}/{{ node.sap_instance.upper() }}{{ instance }}
+
+# HA scenario where ASCS and ERS must share a disk to start/stop the processes as cluster
+{% if node.shared_disk_dev is defined %}
+
 {% if ':' in node.shared_disk_dev %} # This means that the device is a nfs share
 {% set device = node.shared_disk_dev %}
 {% set fstype = netweaver.nfs_version %}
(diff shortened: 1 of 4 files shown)


pillar.example
```diff
@@ -94,12 +96,17 @@ netweaver:
       instance: 00
       root_user: root
       root_password: linux
+      # Set the shared disk used in HA environments. Skip this parameter in non HA environments
       shared_disk_dev: /dev/sbd
+      # Init the shared disk. Only used if a shared disk is provided, not in nfs share cases
       init_shared_disk: True
-      sap_instance: ascs
+      # If a nfs share is used to manage the HA mounting point, like in the cloud
```
smpsgill: maybe this comment and `shared_disk_dev` could be moved after `shared_disk_dev: /dev/sbd ` and start with `Or if a nfs share...`

netweaver/setup/mount.sls
```diff
@@ -7,6 +7,13 @@
 
 {% if node.sap_instance.lower() in ['ascs', 'ers'] %}
 
+create_folder_{{ node.sap_instance.lower() }}_{{ instance_name }}:
+  file.directory:
+    - name: /usr/sap/{{ node.sid.upper() }}/{{ node.sap_instance.upper() }}{{ instance }}
+
+# HA scenario where ASCS and ERS must share a disk to start/stop the processes as cluster
+{% if node.shared_diks_dev is defined %}
```
smpsgill: typo here: should be `shared_disk_dev`

smpsgill (changes requested):
Some minor changes @arbulu89

smpsgill (approved):
Looks good. You can merge it after testing!

PR #50, opened 2020-05-18
Use gcp-vcp-move-ip route names properly

arbulu89:
I found out that we were using the route names incorrectly, using only 1 route for both ASCS and ERS. In my last try crmsh complained with:
```
xarbulu-netweaver01:~ # /usr/sbin/crm configure load update /tmp/cluster.config
WARNING: Resources rsc_ip_HA1_ASCS00,rsc_ip_HA1_ERS10 violate uniqueness for parameter "route_name": "xarbulu-nw-route"
Do you still want to commit (y/n)? n
```

I realized that it makes a total sense, and we need to provide 2 route names.


diegoakechi (approved):
Yes, it makes sense.

PR #51, opened 2020-05-27
Add support to extract nw media archives

smpsgill:
With this PR some new changes:
-Add support to extract NW additional DVDs, and update the dvds list accordingly. Some DVDs can be provided as archive files, some folders or all archives/folders
-Remove the `swpm_extract_dir` variable and replace by `nw_extract_dir`

Now the extracted media will be located in sub directories under the `nw_extract_dir`

Example:
`/sapmedia/NW:                                                                                                 `
`HANA_MEDIA_51053787          NETWEAVER_EXPORT_51050829_3           SWPM`

The subdirectories will be SWPM or dvd names with file extensions removed.

The SWPM installer will look for the already extracted folders or the extract paths based on `nw_extract_dir`

To-Do:
-update the `nw_extract_pydbapi_client` software folders to now include the extracted dvds
-Test with various deployment scenarios
-Update the spec, change files
-Create PR for related terraform project changes


netweaver/extract_nw_archives.sls
```diff
@@ -0,0 +1,51 @@
...
+    - output_dir: {{ nw_extract_dir }}/SWPM
+    - options: "-manifest SIGNATURE.SMF"
+
+{% endif %}
+
+{% set additional_dvd_folders = [] %}
+
+{% for dvd in netweaver.additional_dvds %}
+{% set dvd_folder = dvd.split('.')[0] %}
+{% if additional_dvd_folders.append(dvd_folder) %}{% endif %}
```
arbulu89: This line doesn't make sense, you are running an operation in an `if`. This looks a bad practice for me in this scenario.
What do you want to do here? You could just do `{% do additional_dvd_folders.append(dvd_folder) %}` I guess

netweaver/extract_nw_archives.sls
```diff
@@ -0,0 +1,51 @@
...
+    - sapcar_exe: {{ netweaver.sapcar_exe_file }}
+    - output_dir: {{ nw_extract_dir }}/SWPM
+    - options: "-manifest SIGNATURE.SMF"
+
+{% endif %}
+
+{% set additional_dvd_folders = [] %}
+
+{% for dvd in netweaver.additional_dvds %}
+{% set dvd_folder = dvd.split('.')[0] %}
```
arbulu89: This split looks dangerous, what if the file is `x.tar.gz` or similar. We should check if jinja offers some way to get the path without the extension

netweaver/extract_nw_archives.sls
```diff
@@ -0,0 +1,51 @@
...
+{% if additional_dvd_folders.append(dvd_folder) %}{% endif %}
+
+{%- if dvd.endswith((".ZIP", ".zip", ".RAR", ".rar")) %}
+
+extract_nw_archive:
+  archive.extracted:
+    - name: {{ dvd_folder }}
```
arbulu89: I don't know if we should use `dvd_folder` here.
What if the dvd path is `my_folder/51053787`, this would extract to `/sapmedia/nw/my_folder/51053787`, is this what we want? Or do we want `/sapmedia/nw/51053787`

netweaver/install_aas.sls
```diff
@@ -1,4 +1,6 @@
 {%- from "netweaver/map.jinja" import netweaver with context -%}
+{%- from "netweaver/extract_nw_archives.sls" import additional_dvd_folders with context -%}
+{%- from "netweaver/extract_nw_archives.sls" import swpm_extract_dir with context -%}
```
arbulu89, line 3: I don't see we need this import. We could put `swpm_extract_dir` default value in `defaults.yaml` and avoid all of this imports

netweaver/install_aas.sls
```diff
@@ -63,7 +65,7 @@ netweaver_install_{{ instance_name }}:
     - name: {{ node.sid.lower() }}
     - inst: {{ instance }}
     - password: {{ netweaver.sid_adm_password|default(netweaver.master_password) }}
-    - software_path: {{ netweaver.swpm_folder|default(netweaver.swpm_extract_dir) }}
+    - software_path: {{ netweaver.swpm_folder|default(swpm_extract_dir) }}
```
arbulu89, line 68: I would put the value in `defaults.yaml` and that's it

pillar.example
```diff
@@ -31,18 +31,19 @@ netweaver:
...
   sapexe_folder: your_sapexe_sar_folder_absolute_path
   # Folder where the installation files are stored. /tmp/swpm_unattended by default. Set None to use
   # SAP default folders (it will only work with ASCS and ERS).
   # This folder content will be removed before the installation so be extra careful!
   installation_folder: /tmp/swpm_unattended
   # DB/PAS/AAS instances require additional DVD folders like NW Export or HDB Client folder
-  additional_dvd:
+  # Provide the software folder or archives with additional SAP software needed to install netweaver
+  additional_dvds:
     - required_dvd1_folder_absolute_path
```
arbulu89: We had this `absolute_path` description before, but it looks incorrect, actually the path is relative, right?

netweaver/extract_nw_archives.sls
arbulu89: In fact it's not only `dvd_folder`, it should be `nw_extract_dir~dvd_folder`, isn't it?

arbulu89 (changes requested):
@Simranpal In general it looks good, but I have find out some topics we need to clarify or improve.

netweaver/extract_nw_archives.sls
smpsgill: I used the `{% do additional_dvd_folders.append(dvd_folder) %}` initially, but testing with jinja live parser gave me error using `do`. It seems we need to import the do expression extension. https://jinja.palletsprojects.com/en/master/extensions/#expression-statement

But since we already use it in our formulas and its working fine there I will try to test it now and if its working i will update

netweaver/install_aas.sls
smpsgill: Thanks I will try that

pillar.example
smpsgill: The dvd path in `additional_dvds` is relative to `netweaver_inst_media` in terraform.
In the formula it needs absolute path, so here its correct. I guess thats why you change the automatic pillar adding the '/sapmedia/NW' prefix to dvd pillar entries. I think that should be updated, I will do that in the terraform project PR

netweaver/install_aas.sls
smpsgill: @arbulu89 `swpm_extract_dir` is based on pillar entry : `nw_extract_dir'~'/SWPM`
Defining in `defaults.yaml` sets a fixed value for it

I want to have the nw media extracted to subfolders under a user defined location(`nw_extract_dir`). And use that location as default

netweaver/extract_nw_archives.sls
smpsgill: Yes I checked for alternatives but could not find something in jinja.
We are extracting mainly the rar,exe,zip this is fine for that

smpsgill: Other alternative I found is to use the salt's jinja filter for regex https://docs.saltstack.com/en/latest/topics/jinja/index.html#regex-match

netweaver/extract_nw_archives.sls
smpsgill: Thanks I updated the logic to just fetch the trailing folder name using `salt['file.basename']` instead of including the full relative path

pillar.example
arbulu89: You are right! Thanks for confirming

netweaver/extract_nw_archives.sls
```diff
@@ -0,0 +1,51 @@
...
+    - output_dir: {{ nw_extract_dir }}/SWPM
+    - options: "-manifest SIGNATURE.SMF"
+
+{% endif %}
+
+{% set additional_dvd_folders = [] %}
+
+{% for dvd in netweaver.additional_dvds %}
+{% set dvd_folder = salt['file.basename'](dvd.split('.')[0]) %}
+{% do additional_dvd_folders.append(nw_extract_dir | path_join(dvd_folder)) %}
```
arbulu89: You could set the `nw_extract_dir | path_join(dvd_folder)` part to a variable as it is used 3 times

arbulu89 (approved):
@Simranpal The code looks nice! I have some doubts anyway:

- Is the `additinoal_dvds` import working? This variable is created in a loop, and jinja has a quite strange way to manage the variables scopes, so maybe this variable is not available out of this loop. If it works, I didn't say anything hehe
- Now we have 2 variables to define the SWPM folder/archive: `swpm_folder` and `swpm_sar_file`. Both are self exclusive. I don't know if we could just use 1 for both options, like we do in hana formula. So if the provided path is a folder, we consider that it's already extracted, otherwise it has a `sar` extension, do what you have implemented now. Does it make sense? This could be an improvement for a future PR anyway, but if we can do it, it would be better.

netweaver/extract_nw_archives.sls
```diff
@@ -0,0 +1,51 @@
...
+    - mode: 755
+    - makedirs: True
+
+{% if netweaver.sapcar_exe_file is defined and netweaver.swpm_sar_file is defined %}
+
+extract_installer_file:
+    sapcar.extracted:
+    - name: {{ netweaver.swpm_sar_file }}
+    - sapcar_exe: {{ netweaver.sapcar_exe_file }}
+    - output_dir: {{ nw_extract_dir }}/SWPM
```
arbulu89: You should use `swpm_extract_dir` here I guess

netweaver/extract_nw_archives.sls
smpsgill: Yes I set it to `dvd_extract_dir`

netweaver/extract_nw_archives.sls
```diff
@@ -0,0 +1,58 @@
...
+
+{%- if dvd.endswith((".ZIP", ".zip", ".RAR", ".rar", ".exe", ".EXE")) %}
+{% do additional_dvd_folders.append(dvd_extract_dir) %}
+{%- else %}
+{% do additional_dvd_folders.append(dvd) %}
+{%- endif %}
+
+{%- if dvd.endswith((".ZIP", ".zip", ".RAR", ".rar")) %}
+
+extract_nw_archive:
```
arbulu89: I have noticed that you are using a unmutable state name inside a loop here.
This will create errors when the for loop is executed multiple times (what makes sense).
Use something unique here to extend the state name

netweaver/extract_nw_archives.sls
```diff
@@ -0,0 +1,58 @@
...
+  archive.extracted:
+    - name: {{ dvd_extract_dir }}
+    - enforce_toplevel: False
+    - source: {{ dvd }}
+
+{%- elif dvd.endswith((".exe", ".EXE")) %}
+
+{% set unrar_package = 'unrar_wrapper' if grains['osrelease_info'][0] == 15 else 'unrar' %}
+install_unrar_package:
```
arbulu89: The same here. This will fail in a loop. Actually, you could move this install outside the loop.

netweaver/extract_nw_archives.sls
```diff
@@ -0,0 +1,58 @@
...
+  pkg.installed:
+    - name: {{ unrar_package }}
+
+extract_nw_multipart_archive:
```
arbulu89: Same here

arbulu89 (changes requested):
Hi @Simranpal ,
I have added some more things to improve. About the operations we do in `map.jinja`, you can remove the line where we modify the `additional_dvds` if you need. This was added to fix the issues in older salt versions (SLE12) where `u` character is added to the elements in a list.
This line is just a side effect. I don't mind the solution as long as this works for SLE12 and SLE15. So, make the changes, and test in SLE12 and SLE15 please

Besides that, update the `pillar.example` to have the different extensions in the `additional_dvds` so they are included in the tests.

netweaver/extract_nw_archives.sls
smpsgill: The `install_unrar_package` to install `unrar_wrapper/unrar` is run only when need to extract multi-rar(exe) file extension.
I dont know if it makes sense to install always

arbulu89: Ok, but it must be executed only once. What if you have two `.exe` files? This would file. One option is  execute inside an `if` with `loop.first` condition

netweaver/extract_nw_archives.sls
```diff
@@ -0,0 +1,58 @@
...
+extract_nw_archive_{{ dvd }}:
+  archive.extracted:
+    - name: {{ dvd_extract_dir }}
+    - enforce_toplevel: False
+    - source: {{ dvd }}
+
+{%- elif dvd.endswith((".exe", ".EXE")) %}
+
+{% set unrar_package = 'unrar_wrapper' if grains['osrelease_info'][0] == 15 else 'unrar' %}
+install_unrar_package_to_extract__{{ dvd }}:
```
arbulu89: If you want you could add this in an `if` condition with `loop.first` option.

pillar.example
```diff
@@ -31,20 +31,21 @@ netweaver:
...
   # SAP default folders (it will only work with ASCS and ERS).
   # This folder content will be removed before the installation so be extra careful!
   installation_folder: /tmp/swpm_unattended
   # DB/PAS/AAS instances require additional DVD folders like NW Export or HDB Client folder
-  additional_dvd:
-    - required_dvd1_folder_absolute_path
-    - required_dvd2_folder_absolute_path
+  # Provide the absolute path to software folder or archives with additional SAP software needed to install netweaver
+  additional_dvds:
+    - /sapmedia/HANA/51053492
```
arbulu89, line 47: Please, add more example. `.exe` and `.sar` extension at least

arbulu89 (approved):
@Simranpal It looks good.
I was wondering what you have decided about removing `swpm_folder` or `swpm_sar_file`, one of them as they are self exclusive. If you want, we can close this PR when we can, and just work on that in the future, but don't forget to create a follow up ticket.

**Please, test this in SLE12 and SLE15 options**

smpsgill:
> @Simranpal It looks good.
> I was wondering what you have decided about removing `swpm_folder` or `swpm_sar_file`, one of them as they are self exclusive. If you want, we can close this PR when we can, and just work on that in the future, but don't forget to create a follow up ticket.
> 
> **Please, test this in SLE12 and SLE15 options**

@arbulu89 I will create another PR for removing extraction related variables, as same is needed for hana too.
I will test this PR with SLE 12/15 versions today and update

netweaver/extract_nw_archives.sls
smpsgill: Thanks I added the `loop.first ` option just for the `install_unrar_package_to_extract__{{ dvd }}` state

pillar.example
smpsgill: @arbulu89 I have added the `.exe` example. For the netweaver media, it exists in other archives format except the `.sar` file. I added extraction support for `.sar` SWPM only. Maybe it makes sense to include the extraction of `.sar` type additional dvds. This will cover the hana sar package. What do you think?

smpsgill:
@arbulu89  Tested with the SLES15SP1 on libvirt and worked as expected. I tested providing a custom extract directory for nw media (used 51050829_3.ZIP as additional dvd), and extraction and installation worked fine. There was other unrelated issue with SLES12 version deployment. I will try that again.

pillar.example
smpsgill: I have added now the logic to extract additional_dvds of `.sar` archives type

arbulu89 (approved):
Thanks @Simranpal It looks fine!
Just wait to merge this. We are finishing the submission for SLE15SP2, and this brings non backward compatible changes, so let's finish the submission first.

diegoakechi:
If it is non backward compatible, it shouldn't jump the version only on the minor fix number `0.4.5`. Also I think listing the break changes on the changelog is a good thing to do.

smpsgill:
@arbulu89 I did changes to fix the pydbapi installation, moving this step out of the initial setup, If this looks good, I can merge this PR. I tested with a NW deployment in libvirt

PR #54, opened 2020-07-15
Update the Suse manager form.yml with latest formula data

smpsgill:
Adding entries for the SUMA form based on the latest pillar and formula data

### DIFF
--- a/form.yml
+++ b/form.yml
@@ -17,37 +17,47 @@ netweaver:
           $name: IP Address
       $type: text
       $name: Hostname
-  
+
+  sidadm_user_checkbox:
+    $name: Create the SAP sidadm user and group before the installation
+    $type: boolean
+    $default: false      
   sidadm_user:
     $name: SAP User/Group
-    $type: edit-group
-    $help: Create the SAP sidadm user and group before the installation
-    $prototype:
-      uid:
-        $name: User ID
-        $type: text
-        $optional: true
-      gid:
-        $name: Group ID
-        $type: text
-        $optional: true
-  
-  sap_passwords_checkbox:
-    $name: Provide SAP user passwords
+    $visibleIf: .sidadm_user_checkbox == true
+    $type: group
+    $help: If this entry exists the user and group will be created before the installation, not otherwise
+    uid:
+      $name: User ID
+      $type: text
+    gid:
+      $name: Group ID
+      $type: text
+
+  sid_adm_password_checkbox:
+    $name: Provide SAP <sid>adm user password
     $type: boolean
     $default: false
-  sidadm_password:
+    $help: <sid>adm password is optional, master password will be used as default, if the value is not defined
+  sid_adm_password:
     $name: SAP <sid>adm user password
-    $visibleIf: .sap_passwords_checkbox == true
+    $visibleIf: .sid_adm_password_checkbox == true
     $type: password
-    $help: The password of the SAP <sid>adm user
+    $help: The password of the SAP <sid>adm user. If not set master password will be used as default
     $optional: true
-  sapadm_password:
-    $name: SAP admin password
-    $visibleIf: .sap_passwords_checkbox == true
+  
+  sap_adm_password_checkbox:
+    $name: Provide SAP <sap>adm user password
+    $type: boolean
+    $default: false
+    $help: <sap>adm password is optional, master password will be used as default, if the value is not defined
+  sap_adm_password:
+    $name: SAP admin <sap>adm user password
+    $visibleIf: .sap_adm_password_checkbox == true
     $type: password
-    $help: The password of the SAP <sap>adm user
+    $help: The password of the SAP <sap>adm user. If not set master password will be used as default
     $optional: true
+  
   master_password:
     $name: Master password
     $type: password
--- a/form.yml
+++ b/form.yml
@@ -11,7 +11,7 @@ netweaver:
   virtual_addresses:
     $name: Virtual Addresses
     $type: edit-group
-    $help: Define the IP addresses and hostnames for machines used in SAP Netweaver installation. These machines will be part of the HA cluster. Each entry must contain an IP address and a hostname such as 192.168.201.111 & hacert01
+    $help: Define the virtual IP addresses and virtual hostnames for all of the SAP machines that belong to the same installation. These virtual hostnames are later used internally by SAP to know the other machines in the same landscape. Each entry must contain an IP address and a hostname such as 192.168.201.111 & hacert01
     $prototype:
       $key:
           $type: text
@@ -73,8 +73,8 @@ netweaver:
   clean_nfs:
     $name: Clean NFS share content
     $type: boolean
-    $default: true
-    $help: Clean /sapmnt/{sid} and /usr/sap/{sid}/SYS content. It will only work if ASCS node is defined.
+    $default: false
+    $help: Clean /sapmnt/{sid} and /usr/sap/{sid}/SYS content. It will only work if ASCS node is defined. Note this is destructive operation and will remove the NFS share contents 
     $optional: false
 
   sapmnt_inst_media:
@@ -139,7 +139,7 @@ netweaver:
     $name: NFS version
     $type: text
     $default: nfs4
-    $help: NFS version used to connect to the nfs share
+    $help: NFS version used to connect to the sapmnt nfs share
     $optional: true
 
   nfs_options:
@@ -299,8 +299,9 @@ netweaver:
         $optional: true      
       sap_instance:
         $name: SAP Instance
-        $type: text
+        $type: select
         $help: Provide the netweaver instance/server type - ascs, ers, db, pas, aas
+        $values: [ascs, ers, db, pas, aas]
         $optional: false
       extra_parameters_checkbox:
         $name: Provide extra installation parameters
--- a/form.yml
+++ b/form.yml
@@ -30,9 +30,17 @@ netweaver:
     uid:
       $name: User ID
       $type: text
+      $optional: true
     gid:
       $name: Group ID
       $type: text
+      $optional: true
+
+  master_password:
+    $name: Master password
+    $type: password
+    $help: Master password is used for all the SAP users that are created
+    $optional: false
 
   sid_adm_password_checkbox:
     $name: Provide SAP <sid>adm user password
@@ -57,12 +65,6 @@ netweaver:
     $type: password
     $help: The password of the SAP <sap>adm user. If not set master password will be used as default
     $optional: true
-  
-  master_password:
-    $name: Master password
-    $type: password
-    $help: Master password is used for all the SAP users that are created
-    $optional: false
 
   clean_nfs:
     $name: Clean NFS share content
@@ -92,8 +94,8 @@ netweaver:
     $name: Path to extract the Netweaver installation media
     $visibleIf: .use_swpm_sar_file == true
     $type: text
-    $default: /sapmedia/NW
-    $help: The NW media will be extracted to this path. By default this path is /sapmedia/NW
+    $default: /sapmedia_extract/NW
+    $help: The NW media will be extracted to this path. By default this path is /sapmedia_extract/NW
     $optional: true
   sapcar_exe_file:
     $name: Path to sapcar executable if extracting NW SAR archive
@@ -104,6 +106,7 @@ netweaver:
 
   swpm_folder:
     $name: SWPM folder location
+    $visibleIf: .use_swpm_sar_file == false
     $type: text
     $help: Path to already extracted SWPM installer folder
     $optional: true
@@ -111,21 +114,21 @@ netweaver:
   sapexe_folder:
     $name: SAP exe folder location
     $type: text
-    $help: Path to sap exe folder
+    $help: Path to sap exe folder. This folder must contain all the downloaded software from SAP download center to install the Netweaver systems. like SAPEXE, SAPEXEDB, igshelper, etc
     $optional: false
   
   installation_folder:
     $name: Netweaver installation folder
     $type: text
     $default: /tmp/swpm_unattended
-    $help: Path to run the installation
+    $help: Folder where the installation is executed. It must be an empty folder. The execution logs will be stored there
     $optional: false
   
   ha_enabled:
     $name: Enable HA cluster configuration
     $type: boolean
     $default: true
-    $help: Enable the HA cluster configuration
+    $help: Enable the HA cluster configuration. In order to make this work, ASCS and ERS system must be installed together in parallel.
     $optional: false
 
   nfs_version:
@@ -202,7 +205,7 @@ netweaver:
   product_id:
     $name: SAP product ID
     $type: text
-    $help: The installed SAP product id. If some particular node has its own product_id that one will have the preference
+    $help: The installed SAP product id. If some particular node has its own product_id that one will have the preference. The options are. NW750.HDB.ABAPHA, S4HANA1709.CORE.HDB.ABAPHA, S4HANA1809.CORE.HDB.ABAPHA, S4HANA1909.CORE.HDB.ABAPHA. Remove the 'HA' from the end for non HA environment (without ERS)
     $optional: false
 
   monitoring_enabled:
@@ -263,7 +266,7 @@ netweaver:
         $type: password
         $optional: false      
       shared_disk_dev:
-        $name: Shared disk device
+        $name: Shared disk device (valid only for libvirt provider)
         $type: text
         $optional: true
         $help: Set the shared disk used in HA environments. Skip this parameter in non HA environments
@@ -286,7 +289,7 @@ netweaver:
       sap_instance:
         $name: SAP Instance
         $type: text
-        $help: Provide the netweaver instance/server type
+        $help: Provide the netweaver instance/server type - ascs, ers, db, pas, aas
         $optional: false
       extra_parameters_checkbox:
         $name: Provide extra installation parameters
--- a/form.yml
+++ b/form.yml
@@ -225,6 +225,16 @@ netweaver:
         $name: Virtual Hostname
         $type: text
         $optional: false
+      ascs_virtual_host_checkbox:
+        $name: Provide the ASCS Virtual Hostname (needed for PAS instance only)
+        $type: boolean
+        $default: false
+        $help: If a nfs share is used to manage the HA mounting point
+      ascs_virtual_host:
+        $name: ASCS Virtual Hostname
+        $visibleIf: .ascs_virtual_host_checkbox == true
+        $type: text
+        $optional: true
       virtual_host_interface:
         $name: Virtual host interface
         $type: text
--- a/form.yml
+++ b/form.yml
@@ -268,14 +268,15 @@ netweaver:
         $type: boolean
         $optional: true
         $help: Init the shared disk. Only used if a shared disk is provided
-      node_product_id:
+      product_id:
         $name: SAP product ID
         $type: text
-        $help: The installed SAP product id. This will have the preference over global product id
+        $help: Set an specific product id. This will have the preference over global product id
         $optional: true      
       sap_instance:
         $name: SAP Instance
         $type: text
+        $help: Provide the netweaver instance/server type
         $optional: false
       extra_parameters_checkbox:
         $name: Provide extra installation parameters
@@ -297,17 +298,19 @@ netweaver:
       retry_checkbox:
         $name: Retry the SAP instance installation
         $type: boolean
-        $help: How often retry the SAP instance installation 
+        $help: Retry installation options configuration. Only for ERS, DB, PAS and AAS instances
         $optional: true
-      retry_attempts:
+      attempts:
         $name: Retry attempts
         $visibleIf: .retry_checkbox == true
         $type: text
         $default: 10
+        $help: How often to retry the SAP instance installation. (10 by default retries)
         $optional: true
-      retry_interval:
+      interval:
         $name: Retry interval
         $visibleIf: .retry_checkbox == true
         $type: text
         $default: 60
+        $help: Interval in seconds between retries (60 seconds by default for DB and 600 seconds for PAS and AAS)
         $optional: true
--- a/form.yml
+++ b/form.yml
@@ -11,7 +11,7 @@ netweaver:
   virtual_addresses:
     $name: Virtual Addresses
     $type: edit-group
-    $help: Define the IP addresses and hostnames for machines used in SAP Netweaver installation. These machines will be part of the HA cluster
+    $help: Define the IP addresses and hostnames for machines used in SAP Netweaver installation. These machines will be part of the HA cluster. Each entry must contain an IP address and a hostname such as 192.168.201.111 & hacert01
     $prototype:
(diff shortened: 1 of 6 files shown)


arbulu89 (commented):
@Simranpal I cannot really review just with the code. I think we need to have a SUMA instance available in our network with these changes so we can play with the formulas

arbulu89:
@Simranpal How is the state of this PR?

MalloZup (commented):
if we do have all needed pillar here and we tested that is ok and working we can imho merge this.

Unless we are missing something? :bat:

smpsgill:
> needed pillar here and we teste

a part of testing is left for the netweaver instance installation states, I will update it finishing that part. I have updated form to match other states, and will check if the installation states are run correctly

smpsgill:
@arbulu89 @MalloZup 
I have done the testing for almost of the Netweaver states and the pillar data and states are shown correctly except install state(see detail below 3.)
Some topics/issues I noticed during the testing:

1. form may be too long, and you may not see entries at bottom and may miss the configuration. The workaround is to close the drop-down menus above and zoom-out
2. Testing with setting the same hostname for multiple instances may cause conflicts in some states, and report error in highstate
3. The `netweaver_install_{{ instance_name }}` state for all instances shows strange behaviour when checking the highstate in SUMA web interface. `netweaver.installed:` gets split into `netweaver:` and installed is passed as a list entry to it. See the image attached.
4. a fix is needed for import statements in install sls files(same as hana), I can do that in seperate PR as that is jinja/salt problem.
5. (Same for other forms) Sometimes the form retains old form values, and old states being run, even when reapplying the formula

smpsgill:
Issue with `netweaver.installed:`
image

metadata.yml
```diff
@@ -0,0 +1,4 @@
+description:
+  SAP Netweaver deployment formula
+group:
```
arbulu89, line 3: Maybe we should have the same group for `HANA` and `Netweaver`. What do you think?
Something like `SAP deployment`

netweaver/pre_validation.sls
```diff
@@ -0,0 +1,26 @@
...
+
+{# Check for SWPM archive media checkbox #}
+{% if netweaver.use_swpm_sar_file is defined and netweaver.use_swpm_sar_file == false %}
+    {% do netweaver.pop('swpm_sar_file', none) %}
+{% endif %}
+
+{% for node in netweaver.nodes if node.host == host %}
+
+    {# Check for ascs_virtual_host checkbox #}
+    {% if node.ascs_virtual_host_checkbox is defined and node.ascs_virtual_host_checkbox == false %}
```
arbulu89, line 13: I don't understand this. Why `ascs`?
What happens for the other types (`ers, pas, aas`)?

form.yml
```diff
@@ -0,0 +1,326 @@
...
+    $default: false
+    $help: <sap>adm password is optional, master password will be used as default, if the value is not defined
+  sap_adm_password:
+    $name: SAP admin <sap>adm user password
+    $visibleIf: .sap_adm_password_checkbox == true
+    $type: password
+    $help: The password of the SAP <sap>adm user. If not set master password will be used as default
+    $optional: true
+  
+  master_password:
```
arbulu89: I would put the `master_password` before the rest of the password. I think it follows the flow better.
What do you think?

form.yml
```diff
@@ -0,0 +1,326 @@
...
+          $type: text
+          $name: IP Address
+      $type: text
+      $name: Hostname
+
+  sidadm_user_checkbox:
+    $name: Create the SAP sidadm user and group before the installation
+    $type: boolean
+    $default: false      
+  sidadm_user:
```
arbulu89, line 27: Shouldn't `sidadm_user` be optional?
We don't force to have this entry, that for sure.

form.yml
```diff
@@ -0,0 +1,326 @@
...
+    $name: Path to SWPM archive
+    $visibleIf: .use_swpm_sar_file == true
+    $type: text
+    $help: The path to swpm installer media archive in SAR format. You also need to provide the sapcar executable path for extraction
+    $optional: true
+  nw_extract_dir:
+    $name: Path to extract the Netweaver installation media
+    $visibleIf: .use_swpm_sar_file == true
+    $type: text
+    $default: /sapmedia/NW
```
arbulu89: Change the default to something else. We use `sapmedia_extract/NW` in the terraform project

form.yml
```diff
@@ -0,0 +1,326 @@
...
+    $help: The NW media will be extracted to this path. By default this path is /sapmedia/NW
+    $optional: true
+  sapcar_exe_file:
+    $name: Path to sapcar executable if extracting NW SAR archive
+    $visibleIf: .use_swpm_sar_file == true
+    $type: text
+    $help: The path to sapcar executable to extract NW SAR archive
+    $optional: true
+
+  swpm_folder:
```
arbulu89, line 107: I guess this `swpm_folder` should be self exclusive with the SAR option, right?

form.yml
```diff
@@ -0,0 +1,326 @@
...
+    $name: SWPM folder location
+    $type: text
+    $help: Path to already extracted SWPM installer folder
+    $optional: true
+  
+  sapexe_folder:
+    $name: SAP exe folder location
+    $type: text
+    $help: Path to sap exe folder
```
arbulu89: I would add in the help entry something like `This folder must contain all the downloaded software from SAP download center to install the Netweaver systems. E.g.: SAPEXE, SAPEXEDB, igshelper, etc

form.yml
```diff
@@ -0,0 +1,326 @@
...
+    $optional: false
+  
+  installation_folder:
+    $name: Netweaver installation folder
+    $type: text
+    $default: /tmp/swpm_unattended
+    $help: Path to run the installation
```
arbulu89: I would change the help to `Folder where the installation is executed. It must be an empty folder. The execution logs will be stored there`

form.yml
```diff
@@ -0,0 +1,326 @@
...
+    $optional: false
+  
+  ha_enabled:
+    $name: Enable HA cluster configuration
+    $type: boolean
+    $default: true
+    $help: Enable the HA cluster configuration
```
arbulu89: Add in the help. `In order to make this work, ASCS and ERS system must be installed together in parallel.`

form.yml
```diff
@@ -0,0 +1,326 @@
...
+      $name: Schema name
+      $type: text
+    password:
+      $name: Password
+      $type: password
+
+  product_id:
+    $name: SAP product ID
+    $type: text
+    $help: The installed SAP product id. If some particular node has its own product_id that one will have the preference
```
arbulu89: We need to add more information here. The product ID is not something everyone will know (actually I would be nobody knows, the options are. `NW750.HDB.ABAPHA, S4HANA1709.CORE.HDB.ABAPHA, S4HANA1809.CORE.HDB.ABAPHA, S4HANA1909.CORE.HDB.ABAPHA`. If we remove the `HA` from the end, it means a non HA environment (without ERS)

form.yml
```diff
@@ -0,0 +1,326 @@
...
+        $optional: false
+      root_user:
+        $name: Root user
+        $type: text
+        $optional: false
+      root_password:
+        $name: Root user password
+        $type: password
+        $optional: false      
+      shared_disk_dev:
```
arbulu89, line 298: I think this entry is only for `libvirt`. We need to explain this

form.yml
```diff
@@ -0,0 +1,326 @@
...
+        $help: Init the shared disk. Only used if a shared disk is provided
+      product_id:
+        $name: SAP product ID
+        $type: text
+        $help: Set an specific product id. This will have the preference over global product id
+        $optional: true      
+      sap_instance:
+        $name: SAP Instance
+        $type: text
+        $help: Provide the netweaver instance/server type
```
arbulu89: Put the options here in the help: `ascs, ers, db, pas, aas`

arbulu89 (changes requested):
Hey @Simranpal , Thank you for the PR.
I think that it needs a lot of work, specially documenting the used values. We need to provide a lot of more information. Just think, if we give this to any of our colleagues, would they be able to know what to put?

About the `installed` issue, I don't know, it looks strange. We need to research more, I don't know why it would be rendered wrong.

metadata.yml
smpsgill: yes its a good idea I will update it

netweaver/pre_validation.sls
smpsgill: I checked the installation states, the `ascs_virtual_host` entry is needed for `pas` installation. Is this correct?
https://github.com/SUSE/sapnwbootstrap-formula/blob/84ced2387dfdf481ffab0c8eea37767340ecf7e1/netweaver/install_pas.sls#L32
https://github.com/SUSE/sapnwbootstrap-formula/blob/84ced2387dfdf481ffab0c8eea37767340ecf7e1/pillar.example#L138

So i have asked it in the form. It is also added in pillar.example for db isntance.. but not used for installation? Is the pillar example correct or should the  `ascs_virtual_host` be removed from `db` instance entry?

form.yml
smpsgill: I have followed the layout of pillar data for order of the form. Any way is fine, I can update it ask master password first

form.yml
smpsgill: correct, I have a checkbox for it and it is by default `false`. And since this is `group`, I can remove it via pre validation

form.yml
smpsgill: Yes, this needs to be updated in prevalidation logic, thanks

form.yml
smpsgill: Thanks I will update it

netweaver/pre_validation.sls
arbulu89: Ok, now I understand.

form.yml
smpsgill: @arbulu89 I have added `optional` to elements inside `sidadm_user`. This will cover the case if sidadm_user is true/false. In case of checkbox true and subelements have empty values error will be generated.
If the user leaves the subelements filled, but does not set the checkbox, the state will be run. It is not possible to do the `pop` element operation at root elements of `netweaver`. So this case is same for other checkbox cases under `netweaver`

form.yml
smpsgill: Actually, need to update the visibility, which is self exclusive with SAR options.

The logic to pop elements in netweaver did not work for these self exclusive options

smpsgill:
> Hey @Simranpal , Thank you for the PR.
> I think that it needs a lot of work, specially documenting the used values. We need to provide a lot of more information. Just think, if we give this to any of our colleagues, would they be able to know what to put?
> 
> About the `installed` issue, I don't know, it looks strange. We need to research more, I don't know why it would be rendered wrong.

@arbulu89 I did the changes. Regarding the issue with strange rendering of states, actually it occurs for other states as well. Running the highstate the states are run normally. It is mostly a rendering issue

form.yml
```diff
@@ -0,0 +1,329 @@
...
+        $optional: false
+
+  saptune_solution:
+    $name: saptune solution to apply
+    $type: text
+    $default: NETWEAVER
+    $help: saptune solution to apply to all nodes
+    $optional: true
+
+  hana_checkbox:
```
arbulu89, line 171: We say that is mandatory only for `DB/PAS/AAS` but later all of the are options, right? We should change this

form.yml
```diff
@@ -0,0 +1,329 @@
...
+        $optional: false
+      virtual_host:
+        $name: Virtual Hostname
+        $type: text
+        $optional: false
+      ascs_virtual_host_checkbox:
+        $name: Provide the ASCS Virtual Hostname (needed for PAS instance only)
+        $type: boolean
+        $default: false
+        $help: If a nfs share is used to manage the HA mounting point
```
arbulu89: Is this `help` entry correct?

form.yml
```diff
@@ -0,0 +1,329 @@
...
+    $name: Nodes
+    $type: edit-group
+    $minItems: 1
+    $itemName: ""
+    $prototype:
+      host:
+        $name: Hostname to install Netweaver
+        $type: text
+        $optional: false
+      virtual_host:
```
arbulu89, line 240: We have to make clear here, that this `ascs_virtual_host` must match with one of the entries we have added at the top in `virtual_hosts`

form.yml
```diff
@@ -0,0 +1,329 @@
...
+      virtual_host_mask:
+        $name: Virtual host mask
+        $type: text
+        $default: 32
+        $optional: true
+      sid:
+        $name: SAP system identifier (SID)
+        $type: text
+        $optional: false
+        $help: System ID of the HANA installation
```
arbulu89: Maybe add in the `help` that it mast be a 3 characters string and add some examples like `PRD`, `HA1`, etc

form.yml
```diff
@@ -0,0 +1,329 @@
...
+      instance:
```
arbulu89, line 283: Add help please, explain that it must be a 2 digits number

form.yml
```diff
@@ -0,0 +1,329 @@
...
+        $name: SAP instance number
+        $type: text
+        $optional: false
+      root_user:
+        $name: Root user
+        $type: text
+        $optional: false
+      root_password:
```
arbulu89, line 293: Explain that if the `root` user is passwordless, this value can be anything

form.yml
```diff
@@ -0,0 +1,329 @@
...
+        $help: Provide the netweaver instance/server type - ascs, ers, db, pas, aas
+        $optional: false
+      extra_parameters_checkbox:
+        $name: Provide extra installation parameters
+        $type: boolean
+        $optional: true  
+      extra_parameters:
+        $name: Installation extra options
+        $visibleIf: .extra_parameters_checkbox == true
+        $help: Optional configuration parameters
```
arbulu89: Please, improve the help. The extra parameters are additional `key/value` entries that will be updated in the used installation configuration file

form.yml
```diff
@@ -0,0 +1,329 @@
...
+  $name: Netweaver
+  $type: group
+  
+  install_packages:
+    $name: Install required packages
+    $type: boolean
+    $default: true
+    $help: Install all required packages to install SAP Netweaver
+  
+  virtual_addresses:
```
arbulu89, line 11: Add a `help` entry. Explain that this is a `key/value` list or map with the IP addresses and hostnames that later will be used in the installation. That it must contain all the values that will belong to the cluster

arbulu89 (changes requested):
@Simranpal I think there is still a lot of work in the `help` entries.

@MalloZup Could you read all the entries here, and say if you understand what they ask or not? The entries without documentation are really difficult to fill

MalloZup:
yes I think  we need to document the entries. I will try to read the entries from a suse-manager server so I can be on the UX perspective there

MalloZup:
@arbulu89  I think  besied your comments, I think in order to do an atomic change, I would merge this PR where we update the pillars to the latest development.

Maybe @Simranpal  can fix all the remarks.

Once we have this merged, I will re-look in all our formulas documentation and we can check if we need improvements as 2nd iteration.
This is for sure something we will need to do .

form.yml
smpsgill: @arbulu89 I did not understand. Do you mean that this entry should not appear for `ERS/ASCS` instances? In that case, we can detect what type of instance user enters, unless we provide a drop-down option to user to choose the instance type. With this drop-down we can also show/hide some section based on instance type selection.

or you mean we need to set the `$optional: true` for hana sub entries

form.yml
smpsgill: No.. I will update the correct help description

smpsgill:
thanks @arbulu89 @MalloZup .I have updated further the help entries

form.yml
```diff
@@ -0,0 +1,341 @@
...
+  install_packages:
+    $name: Install required packages
+    $type: boolean
+    $default: true
+    $help: Install all the required packages to install SAP Netweaver
+  
+  virtual_addresses:
+    $name: Virtual Addresses
+    $type: edit-group
+    $help: Define the IP addresses and hostnames for machines used in SAP Netweaver installation. These machines will be part of the HA cluster
```
MalloZup: I think we could even provide an example

form.yml
```diff
@@ -0,0 +1,341 @@
...
+  sidadm_user_checkbox:
+    $name: Create the SAP sidadm user and sapsys group before the installation
+    $type: boolean
+    $help: Choose this option if the SAP sidadm user and sapsys group need to be created before the installation
+    $default: false      
+  sidadm_user:
+    $name: SAP User/Group
+    $visibleIf: .sidadm_user_checkbox == true
+    $type: group
+    $help: If this entry exists the user and group will be created before the installation, not otherwise
```
MalloZup: `If this entry exists the user and group will be created before the installation, not during the installation [default]`

form.yml
```diff
@@ -0,0 +1,341 @@
...
+    uid:
+      $name: User ID
+      $type: text
+      $help: User ID for the SAP sidadm user
+      $optional: true
+    gid:
+      $name: Group ID
+      $type: text
+      $help: Group ID for the SAP sidadm user's 'sapsys' group
```
MalloZup: we need to be consistent with our help message.

we should use everywhere: `<sid>adm` because this is the default by SAP doc.

Right now, we are using `sidadm` and later in other `help` `<sid>adm` see line 53.

I do a comment here but everywhere we need to rename from `sidadm`and related  to `<sid>adm`

form.yml
```diff
@@ -0,0 +1,341 @@
...
+    $type: boolean
+    $default: true
+    $help: Enable the HA cluster configuration. In order to make this work, ASCS and ERS system must be installed together in parallel.
+    $optional: false
+
+  nfs_version:
+    $name: NFS version
+    $type: text
+    $default: nfs4
+    $help: Used to connect to the nfs share
```
MalloZup: I would use 
`NFS version used to connect to the nfs share`

MalloZup (changes requested):
some remarks on the help messages. 

I will re-review it during the day

form.yml
smpsgill: thanks I updated it to` <sid>adm`

form.yml
smpsgill: When filling this virtual addresses section in the form, there is IP address and hostname section which is quite straightforward, but I will add an example as you say

form.yml
```diff
@@ -0,0 +1,341 @@
...
+  install_packages:
+    $name: Install required packages
+    $type: boolean
+    $default: true
+    $help: Install all the required packages to install SAP Netweaver
+  
+  virtual_addresses:
+    $name: Virtual Addresses
+    $type: edit-group
+    $help: Define the IP addresses and hostnames for machines used in SAP Netweaver installation. These machines will be part of the HA cluster. Each entry must contain an IP address and a hostname such as 192.168.201.111 & hacert01
```
arbulu89: The part of `These machines will be part of the HA cluster` is not totally correct. The virtual addresses actually are requested by SAP installation, is not something asked by the cluster. Later on, the cluster takes advantage of this addresses, that's true. The machines that don't belong to to the HA cluster, will have the virtual address too (and they must have it).
Documenting this will be super difficult, because depending on the cloud providers, there are some more restrictions (like in AWS for example, where the virtual addresses must be in other different IP range).
I would rewrite to something like:
```
Define the virtual IP addresses and virtual hostnames for all of the SAP machines that belong to the same installation. These virtual hostnames are later used internally by SAP to know the other machines in the same landscape
```
```

File: form.yml [hunk 26]
```diff
@@ -0,0 +1,341 @@
...
+    $default: false      
+  sidadm_user:
+    $name: SAP User/Group
+    $visibleIf: .sidadm_user_checkbox == true
+    $type: group
+    $help: If this entry exists the user and group will be created before the installation, not during the installation [default]
+    uid:
+      $name: User ID
+      $type: text
+      $help: User ID for the SAP <sid>adm user
```
arbulu89: I think there are some restrictions here. Like the user id number must be higher than 999. But I don't know if we will be able to define all of these things, even if we don't know them.

form.yml
```diff
@@ -0,0 +1,341 @@
...
+    $visibleIf: .sap_adm_password_checkbox == true
+    $type: password
+    $help: The password of the SAP <sap>adm user. If not set master password will be used as default
+    $optional: true
+
+  clean_nfs:
+    $name: Clean NFS share content
+    $type: boolean
+    $default: true
+    $help: Clean /sapmnt/{sid} and /usr/sap/{sid}/SYS content. It will only work if ASCS node is defined.
```
arbulu89: Maybe we need to add a remark here saying that this operation is destructive. I would actually put `false` the default option (at least here in the form, where the execution might not be executed in parallel)

form.yml
```diff
@@ -0,0 +1,341 @@
...
+    $type: boolean
+    $default: true
+    $help: Enable the HA cluster configuration. In order to make this work, ASCS and ERS system must be installed together in parallel.
+    $optional: false
+
+  nfs_version:
+    $name: NFS version
+    $type: text
+    $default: nfs4
+    $help: NFS version used to connect to the nfs share
```
arbulu89: Maybe improve adding `... to connect to the sapmnt nfs share`

form.yml
arbulu89: Both hehe
I mean that these entries are only required if you are installing `DB/PAS/AAS`  entries. For `ASCS/ERS` these entries can be spared

form.yml
```diff
@@ -0,0 +1,341 @@
...
+        $help: Init the shared disk. Only used if a shared disk is provided
+      product_id:
+        $name: SAP product ID
+        $type: text
+        $help: Set an specific product id. This will have the preference over global product id
+        $optional: true      
+      sap_instance:
+        $name: SAP Instance
+        $type: text
+        $help: Provide the netweaver instance/server type - ascs, ers, db, pas, aas
```
arbulu89: I would add a `dropdown` option here. These values are managed only by us. We can avoid users putting invalid values

arbulu89 (changes requested):
@Simranpal Some more comments about the `help` entries. I think we are pretty close to have a nice form

form.yml
smpsgill: thanks, yes that would be better description

form.yml
smpsgill: @arbulu89  Yes, I got it. In the form, I think we can implement the logic via drop-down. But the installation of `ASCS/ERS` or `DB/PAS/AAS`  is done at a later step inside the node entries. 

I played around with the `$visibleIf ` option with `some_group#another_group#my_checkbox == true` this shows/hide based on checkbox entries but not on dropdown values.

form.yml
smpsgill: Done, thanks, I tested the installations states of instances like ascs, ers.. after choosing from the dropdown option

form.yml
MalloZup: `User ID for the SAP <sid>adm user (must be higher then 999)`

MalloZup (commented):
thx

arbulu89 (approved):
@Simranpal It looks much better! Great job

MalloZup (approved):

PR #55, opened 2020-08-20
Fix/Update package with pillar.example

smpsgill:
Updating OBS packaging readme, add pillar.example to package and updating the service file version. In previous PR, I forgot to update the service file version


MalloZup (approved):
thx

PR #60, opened 2020-09-14
Update cluster properties in the template

smpsgill:
Some changes to the cluster properties w.r.t Suse best practise guide as suggested by Peter Schinagl. In issue: https://github.com/SUSE/sapnwbootstrap-formula/issues/61

- Move properties `cib-bootstrap-options` ,` rsc_defaults` , `op_defaults` to general instead of AWS specific
- Change to `op monitor interval=11` for ASCS, ERS primitive
- Remove ERS meta resource stickiness

https://documentation.suse.com/sbp/all/single-html/SAP_NW740_SLE15_SetupGuide/

### DIFF
--- a/templates/cluster_resources.j2
+++ b/templates/cluster_resources.j2
@@ -15,8 +15,6 @@
 
 # Platform dependant (stonith, virtual ip address, cib options, etc) resource
 
-{%- if cloud_provider == "amazon-web-services" %}
-
 property cib-bootstrap-options: \
   stonith-enabled="true" \
   stonith-action="off" \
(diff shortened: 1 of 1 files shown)


arbulu89 (commented):
@Simranpal Thank you for the first implementation. I still have doubts if these values are correct for all the providers or not. This is something we need to know beforehand.

arbulu89:
Adding @petersatsuse to the PR too

MalloZup (changes requested):
this can wait SUMA MU .

after https://github.com/SUSE/spacewalk/issues/12421#issuecomment-706292904

yeoldegrove:
outdated by now

PR #63, opened 2020-09-24
Feature/implement ensa differences

arbulu89:
Implement the changes to configure a correct ENSA2 cluster.

1. Identify current ENSA installed version
2. Configure `keepalive` options
3. Update SAP profile files
4. Use the cluster resources template correctly depending on the ENSA version

Based on: 
https://github.com/SUSE/shaptools/pull/59
https://github.com/SUSE/salt-shaptools/pull/72

More information in:
https://www.suse.com/c/suse-ha-sap-certified/
https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse#2d6008b0-685d-426c-b59e-6cd281fd45d7
https://launchpad.support.sap.com/#/notes/1410736

FYI @petersatsuse

### DIFF
--- a/netweaver/ha_cluster.sls
+++ b/netweaver/ha_cluster.sls
@@ -89,7 +89,7 @@ set_keepalive_option_{{ instance_name }}:
     - content: enque/encni/set_so_keepalive = true
     # onlyif statements can be improved when salt version 3000 is used
     # https://docs.saltstack.com/en/latest/ref/states/requisites.html#onlyif
-    - onlyif: cat /etc/salt/grains | grep "ensa_version:.*1"
+    - onlyif: cat /etc/salt/grains | grep "ensa_version_{{ node.sid.lower() }}_{{ instance }}:.*1"
 
 {% elif node.sap_instance.lower() == 'ers' %}
 
--- a/netweaver/ha_cluster.sls
+++ b/netweaver/ha_cluster.sls
@@ -84,7 +84,8 @@ adapt_sap_profile_ascs_{{ instance_name }}:
 set_keepalive_option_{{ instance_name }}:
   file.line:
     - name: {{ profile_file }}
-    - mode: ensure
+    - mode: insert
+    - location: end
     - content: enque/encni/set_so_keepalive = true
     # onlyif statements can be improved when salt version 3000 is used
     # https://docs.saltstack.com/en/latest/ref/states/requisites.html#onlyif
--- a/netweaver/ha_cluster.sls
+++ b/netweaver/ha_cluster.sls
@@ -81,6 +81,15 @@ adapt_sap_profile_ascs_{{ instance_name }}:
     - pattern: '^Restart_Program_01 = local \$\(_EN\) pf=\$\(_PF\)'
     - repl: 'Start_Program_01 = local $(_EN) pf=$(_PF)'
 
+set_keepalive_option_{{ instance_name }}:
+  file.line:
+    - name: {{ profile_file }}
+    - mode: ensure
+    - content: enque/encni/set_so_keepalive = true
+    # onlyif statements can be improved when salt version 3000 is used
+    # https://docs.saltstack.com/en/latest/ref/states/requisites.html#onlyif
+    - onlyif: cat /etc/salt/grains | grep "ensa_version:.*1"
+
 {% elif node.sap_instance.lower() == 'ers' %}
 
 adapt_sap_profile_ers_{{ instance_name }}:
@@ -88,6 +97,12 @@ adapt_sap_profile_ers_{{ instance_name }}:
     - name: {{ profile_file }}
     - pattern: '^Restart_Program_00 = local \$\(_ER\) pf=\$\(_PFL\) NR=\$\(SCSID\)'
     - repl: 'Start_Program_00 = local $(_ER) pf=$(_PFL) NR=$(SCSID)'
+
+remove_autostart_option_{{ instance_name }}:
+  file.line:
+    - name: {{ profile_file }}
+    - match: ^Autostart = 1.*$
+    - mode: delete
 
 {% endif %}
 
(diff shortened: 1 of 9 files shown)


netweaver/setup/keepalive.sls
```diff
@@ -0,0 +1,12 @@
+{%- from "netweaver/map.jinja" import netweaver with context -%}
+
+# Based on https://launchpad.support.sap.com/#/notes/1410736
+# And https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse#2d6008b0-685d-426c-b59e-6cd281fd45d7
+
+{% for sysctl_key, sysctl_value in netweaver.sysctl_values.items()  %}
+
+{{ sysctl_key }}:
+  sysctl.present:
```
Simranpal, line 9: As per the SAP note, we need to pass the `-w` flag when setting the new values. But I could see the salt fn supporting other parameters

templates/cluster_resources.j2
```diff
@@ -1,6 +1,7 @@
 {%- set data = pillar.cluster.configure.template.parameters %}
 {%- set cloud_provider = grains['cloud_provider'] %}
 {%- set sid = data.sid.upper() %}
+{%- set sid_lower = sid.lower() %}
```
smpsgill, line 4: should it not be `data.sid.lower()`

netweaver/setup/keepalive.sls
smpsgill, line 9: As per the SAP note, we need to pass the -w flag when setting the new values. But I could not find the salt fn support for passing addition parameters to sysctl

templates/cluster_resources.j2
```diff
@@ -127,8 +138,7 @@ primitive rsc_sap_{{ sid }}_ASCS{{ ascs_instance }} SAPInstance \
   params InstanceName={{ sid }}_ASCS{{ ascs_instance }}_{{ ascs_virtual_host }} \
      START_PROFILE="/sapmnt/{{ sid }}/profile/{{ sid }}_ASCS{{ ascs_instance }}_{{ ascs_virtual_host }}" \
      AUTOMATIC_RECOVER=false \
-  meta resource-stickiness=5000 failure-timeout=60 \
-     migration-threshold=1 priority=10
+  meta resource-stickiness=5000 {% if ensa_version == 1 %}failure-timeout=60 migration-threshold=1 priority=10{%- endif %}
```
smpsgill, line 141: just wondering, if the `==` comparison works for jinja integer type

netweaver/ha_cluster.sls
```diff
@@ -89,6 +99,12 @@ adapt_sap_profile_ers_{{ instance_name }}:
     - pattern: '^Restart_Program_00 = local \$\(_ER\) pf=\$\(_PFL\) NR=\$\(SCSID\)'
     - repl: 'Start_Program_00 = local $(_ER) pf=$(_PFL) NR=$(SCSID)'
 
+remove_autostart_option_{{ instance_name }}:
+  file.line:
+    - name: {{ profile_file }}
+    - match: ^Autostart = 1.*$
```
smpsgill, line 105: is the` $ ` at end part of regex for searching `Autostart = 1`

netweaver/ha_cluster.sls
```diff
@@ -81,6 +81,16 @@ adapt_sap_profile_ascs_{{ instance_name }}:
...
 
+set_keepalive_option_{{ instance_name }}:
+  file.line:
+    - name: {{ profile_file }}
+    - mode: insert
+    - location: end
+    - content: enque/encni/set_so_keepalive = true
+    # onlyif statements can be improved when salt version 3000 is used
+    # https://docs.saltstack.com/en/latest/ref/states/requisites.html#onlyif
+    - onlyif: cat /etc/salt/grains | grep "ensa_version_{{ node.sid.lower() }}_{{ instance }}:.*1"
```
smpsgill, line 92: We can get version from pillar as you used in other file, but this seems good too.
And does the line in content: `enque/encni/set_so_keepalive = true` needs qoutes?

smpsgill (changes requested):
@arbulu89 one correction about the `data.sid.lower()` and some points to discuss below:

templates/cluster_resources.j2
arbulu89: It's the the same, check the `sid` is initialized just one line before.

netweaver/setup/keepalive.sls
arbulu89: `-w` means write, and the `sysctl.present` does exactly this. Check the value, and write if it's different, so this flag is used correctly

templates/cluster_resources.j2
arbulu89: Yes, it works. You can do tests in http://jinja.quantprogramming.com/ for example

netweaver/ha_cluster.sls
arbulu89: Yes. It's true that it doesn't have many sense, as we have the `.*` before in any case. It means that no more characters should be after this pattern

netweaver/ha_cluster.sls
arbulu89: No, we don't need quotes. At least following the best practices guide.

arbulu89:
@Simranpal Why have you set to request changes? I don't see any requested change hehe, only questions. In any case, all the questions answered

smpsgill:
> @Simranpal Why have you set to request changes? I don't see any requested change hehe, only questions. In any case, all the questions answered

@arbulu89 my bad, I requested changes for the `sid.lower()` but it was correct already. I will approve now

smpsgill (approved):
Thanks for the doubts. looks good!

templates/cluster_resources.j2
smpsgill: Yes i skipped that, thanks

PR #65, opened 2020-09-30
Feature/native fencing

arbulu89:
Enable native fencing optionally.
At some cases, sbd is used as fencing mechanism, and for those cases, we need to disable the native fencing mechanism.
By now, we only have it for gcp and aws (for azure will be the same).

Besides that, I have added the `instance_name` sufix to the socat resources, to unify the naming

FYI: @peteratsuse


smpsgill (approved):
Thanks looks good

MalloZup (approved):

PR #67, opened 2020-10-15
Feature/s4hana

arbulu89:
Add new logic in the database installation template to work properly for S/4HANA.

During S/4HANA installation, the created schema name always is `SAPHANADB`. If you try to use any other name, the installation fails saying that this name cannot be changed.

### DIFF
--- a/templates/db.inifile.params.j2
+++ b/templates/db.inifile.params.j2
@@ -17,7 +17,8 @@
 # HDB_Schema_Check_Dialogs.dropSchema = false
 
 # The name of the database schema.
-{% if schema_name == 'SAPABAP1' %}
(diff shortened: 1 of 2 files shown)


MalloZup (approved):
thx

templates/db.inifile.params.j2
```diff
@@ -17,7 +17,8 @@
 # HDB_Schema_Check_Dialogs.dropSchema = false
 
 # The name of the database schema.
-{% if schema_name == 'SAPABAP1' %}
+# For S/4HANA new versions (1809 and beyond), the DB is always named `SAPHANADB` always
```
MalloZup: good comment

PR #68, opened 2020-11-19
Fix additional_dvds variable usage when salt uses python 2

arbulu89:
The `additional_dvds` management still fails if salt is using python 2 (what happens in SLE12).

The error is:
```OSError: [Errno 2] No such file or directory: "u'/sapmedia/NW/51050829_3'"```
The `u` character is added as prefix in the lists.

This change uses the `tojson` filter to sanitize the string

Additional info about the same issue: https://github.com/saltstack/salt/issues/47001


MalloZup (approved):
thx

PR #69, opened 2020-11-25
Add require statements for nw installation to subsequent states

smpsgill:
If the SAP Netweaver installer fails, the salt states following do not need to be executed. I have added the require statements to some states after netweaver installation, and this will stop further salt states from execution.
I have tested with netweaver deployment and the subsequent states after netweaver instance installation fail as expected.

To abort or reduce the overall runtime of netweaver installation further, incase of any instance installation error, more changes are required.
As sap netweaver instance machines are provisioned in any order and there is dependence between the instances, it can be complex and we need to identify and add more requisites between these instance states and check the feasibility.

### DIFF
--- a/netweaver/install_aas.sls
+++ b/netweaver/install_aas.sls
@@ -87,5 +87,6 @@ remove_aas_inifile_{{ instance_name }}:
     - name: {{ inifile }}
     - require:
       - create_aas_inifile_{{ instance_name }}
+      - netweaver_install_{{ instance_name }}
 
 {% endfor %}
--- a/netweaver/install_aas.sls
+++ b/netweaver/install_aas.sls
@@ -87,6 +87,5 @@ remove_aas_inifile_{{ instance_name }}:
     - name: {{ inifile }}
     - require:
       - create_aas_inifile_{{ instance_name }}
-      - netweaver_install_{{ instance_name }}
 
 {% endfor %}
--- a/netweaver/install_ascs.sls
+++ b/netweaver/install_ascs.sls
@@ -60,5 +60,6 @@ remove_ascs_inifile_{{ instance_name }}:
     - name: {{ inifile }}
     - require:
       - create_ascs_inifile_{{ instance_name }}
+      - netweaver_install_{{ instance_name }}
 
 {% endfor %}
--- a/netweaver/install_ascs.sls
+++ b/netweaver/install_ascs.sls
@@ -60,6 +60,5 @@ remove_ascs_inifile_{{ instance_name }}:
     - name: {{ inifile }}
     - require:
       - create_ascs_inifile_{{ instance_name }}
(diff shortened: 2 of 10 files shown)


netweaver/install_aas.sls
```diff
@@ -87,5 +87,6 @@ remove_aas_inifile_{{ instance_name }}:
     - name: {{ inifile }}
     - require:
       - create_aas_inifile_{{ instance_name }}
+      - netweaver_install_{{ instance_name }}
```
arbulu89: I think we shouldn't put this `require`. We want to remove the configuration file always, not only when the installation works. The config file stores some secrets that we don't want to keep there.

netweaver/install_ascs.sls
```diff
@@ -60,5 +60,6 @@ remove_ascs_inifile_{{ instance_name }}:
     - name: {{ inifile }}
     - require:
       - create_ascs_inifile_{{ instance_name }}
+      - netweaver_install_{{ instance_name }}
```
arbulu89: Same as AAS

netweaver/install_db.sls
```diff
@@ -89,5 +89,6 @@ remove_db_inifile_{{ instance_name }}:
     - name: {{ inifile }}
     - require:
       - create_db_inifile_{{ instance_name }}
+      - netweaver_install_{{ instance_name }}
```
arbulu89: Same as AAS

netweaver/install_ers.sls
```diff
@@ -71,5 +71,6 @@ remove_ers_inifile_{{ instance_name }}:
     - name: {{ inifile }}
     - require:
       - create_ers_inifile_{{ instance_name }}
+      - netweaver_install_{{ instance_name }}
```
arbulu89: Same as AAS

netweaver/install_pas.sls
```diff
@@ -89,5 +89,6 @@ remove_pas_inifile_{{ instance_name }}:
     - name: {{ inifile }}
     - require:
       - create_pas_inifile_{{ instance_name }}
+      - netweaver_install_{{ instance_name }}
```
arbulu89: Same as AAS

arbulu89 (approved):
@Simranpal It looks good in general.
Have a look on my comments to see if they make sense. If so, just remove that require and merge the changes

netweaver/install_aas.sls
smpsgill: I was thinking incase of failure, the inifile maybe useful for debugging.
But yes, from security perspective, that makes sense.

smpsgill:
> @Simranpal It looks good in general.
> Have a look on my comments to see if they make sense. If so, just remove that require and merge the changes

I have removed the nw installation require statement.  I have tested with deployment earlier, so I will merge these changes.

PR #70, opened 2020-12-02
Start the saptune daemon service

smpsgill:
Start the `saptune` daemon on netweaver nodes 'tuned.service' (in saptune version 2) and 'saptune.service' (saptune version 3).
It is needed to get the sap tuning back especially after system reboot.
I have added the saptune daemon start cmd. Tested with Netweaver deployment


MalloZup (approved):
thx

arbulu89 (approved):

PR #72, opened 2021-01-28
Fix error about missing instance installation requisite

smpsgill:
When monitoring is enabled, one of the monitoring state `sap_host_exporter_configuration_{{ exporter_instance }}` is run for all nodes. This cause the error about some missing nw installation states which are part of requisite for the monitoring state.

I have added check for host,to make sure current node installation is checked instead of all nodes.
I have updated the formula version and created related bugzilla entry to submit to SLE.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,12 @@ sudo: required
 services:
   - docker
 
+dist: bionic
 addons:
   apt:
     sources:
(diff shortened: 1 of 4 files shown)


arbulu89 (approved):
@Simranpal Thank you.
I think this will fix the error.
Please, test it properly and see if the exporters are working (even in the case of a failover of the cluster, to see that they are moved properly to the correct node)

Issue #74, opened 2021-02-05
The formula could not be used without a NFS-share


petersatsuse:
As in a distributed sap architecture and with HA we use a NFS-share for the NW profiles etc.
If the formula is not used in the above context, like single node/dual node deployment, the NFS part in the pillar and in the formula need to be manual deleted/uncommented before the deployment can start
It would be helpful to have a switch for such a usage without a nfs-share

arbulu89:
Fixed in: https://github.com/SUSE/sapnwbootstrap-formula/pull/79

PR #79, opened 2021-02-15
Feature/make sapmnt configurable

arbulu89:
Fix for: https://github.com/SUSE/sapnwbootstrap-formula/issues/74

2 things implemented:
1. Install `/sapmnt` locally instead of a NFS server. This can be used in simple deployments where ASCS and PAS are in the same machine and we don't have any HA capability. This is achieved using the `sap_inst_media`. If not used or set to empty string the local option is used. Otherwise it will work as before connecting to a NFS share
2. Provide the way to configure the `sapmnt` folder. By default is uses `/sapmnt` (which is the default value used by SAP). This option can be configured with the `sapmnt_path` pillar option

### DIFF
--- a/netweaver/defaults.yaml
+++ b/netweaver/defaults.yaml
@@ -5,6 +5,7 @@ netweaver:
   nw_extract_dir: /sapmedia_extract/NW
   additional_dvds: []
   ha_enabled: True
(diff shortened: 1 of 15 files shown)


stefanotorresi (approved):
LGTM!

PR #80, opened 2021-02-19
Add some S4HANA specific mandatory values in templates

arbulu89:
Add some mandatory parameters to the NW installation templates for S4HANA.

Until now, they were being added in the `ha-sap-terraform-deployments` project. This is really hard to manage and document, and as the parameters are mandatory they are now in the template.

https://github.com/SUSE/ha-sap-terraform-deployments/blob/develop/pillar_examples/automatic/netweaver/netweaver.sls#L137

**Disclaimer**: I don't backup these changes in the any documentation. This is based in long reverse engineering that I did in the past to know these parameters and values, and this PR just carry overs them here


stefanotorresi (approved):
LGTM

Issue #73, opened 2021-02-03
virtual ip adresses used in the pillar are not persistent


petersatsuse:
if the formula is used standalone and without a cluster, the virtual ip addresses used for the different sap instances (as,er,db,pas,aas, ...) are gone after a reboot.

They are not persistence as they are not written into the /etc/icfg- file.

They simply need to be append to the /etc/sysconfig/network/ifcfg-ethX file like
 IPADDR_1=192.168.201.115/24 
 IPADDR_2=192.168.201.116/24 
 IPADDR_3=192.168.201.117/24 
 IPADDR_4=192.168.201.118/24 
...

arbulu89:
@stefanotorresi FYI

arbulu89:
@petersatsuse How can be reproduce the error created by this and check if the potential fix solves it?
Do you check if the SAP instances are working after a reboot? Do we need to run certain command or check logs?

petersatsuse:
To reproduce, simply install a NW without a cluster. 

The formula creates somehow the virtual interfaces, so the install runs. 
But if you reboot the system, you SAP system can not start anymore as there are no network interfaces for the virtual ip's.

In a cluster environment the virtual ip's are generated and controlled by the cluster-stack and therefore everything is ok.
Using it without a cluster, only work until you reboot the first time.

I proofed that adding the virt. ip to the ifcfg file makes them persistent, and i am able to start the sap system (what was not possible before)

arbulu89:
Fixed by: https://github.com/SUSE/sapnwbootstrap-formula/pull/81

PR #81, opened 2021-02-23
Set permanentnly the virtual addreses for non HA scenarios

arbulu89:
Fix for: https://github.com/SUSE/sapnwbootstrap-formula/issues/73

Now the virtual ip addresses are set permanently in the `/etc/sysconfig/network/ifcfg-ethX` (the network interface is based on the pillar file).

If the scenario is HA, they are removed later on as the virtual address is managed by the cluster and it cannot be permanent.
In this case, the data is remove from the file. This `add first/remove later` looks better as it will remove the address even if some data is not considered in the pillar file, so it covers more scenarios.

Consider that more than 1 virtual addresses can be used in 1 machine (ASCS and PAS in the same machine for example), that's why we need to loop the data many times.

The `STARTMODE` and `BOOTPROTO` are only changed if they don't have any value (these values are needed to make the network interface permanent, if there are not other configuration values like `dhcp`, etc)

### DIFF
--- a/netweaver/ha_cluster.sls
+++ b/netweaver/ha_cluster.sls
@@ -7,6 +7,17 @@
 {% set instance_name = node.sid~'_'~instance %}
 {% set instance_folder = node.sap_instance.upper()~instance %}
 {% set profile_file = '/usr/sap/'~node.sid.upper()~'/SYS/profile/'~node.sid.upper()~'_'~instance_folder~'_'~node.virtual_host %}
+{% set virtual_host_interface = node.virtual_host_interface|default('eth0') %}
+{% set ifcfg_file = '/etc/sysconfig/network/ifcfg-'~virtual_host_interface %}
+
+{% for virtual_ip, hostname in netweaver.virtual_addresses.items() if hostname == node.virtual_host %}
+# Remove permanent ip address as the element is managed by the cluster
+remove_permanent_ipaddr_{{ instance_name }}:
+  file.line:
+    - name: {{ ifcfg_file }}
+    - match: {{ virtual_ip }}
+    - mode: delete
+{% endfor %}
 
 install_suse_connector:
   pkg.installed:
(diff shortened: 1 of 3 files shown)


netweaver/ha_cluster.sls
```diff
@@ -7,6 +7,17 @@
 {% set instance_name = node.sid~'_'~instance %}
 {% set instance_folder = node.sap_instance.upper()~instance %}
 {% set profile_file = '/usr/sap/'~node.sid.upper()~'/SYS/profile/'~node.sid.upper()~'_'~instance_folder~'_'~node.virtual_host %}
+{% set virtual_host_interface = node.virtual_host_interface|default('eth0') %}
+{% set ifcfg_file = '/etc/sysconfig/network/ifcfg-'+virtual_host_interface %}
```
smpsgill: probably better to use the `~` instead of `+`

netweaver/setup/virtual_addresses.sls
```diff
@@ -7,3 +8,40 @@
...
       - {{ hostname }}
 {% endfor %}
+
+
+{% for node in netweaver.nodes if node.host == host %}
+# Set the virtual ip addresses permanently
+{% set instance = '{:0>2}'.format(node.instance) %}
+{% set instance_name = node.sid~'_'~instance %}
+{% set virtual_host_interface = node.virtual_host_interface|default('eth0') %}
+{% set ifcfg_file = '/etc/sysconfig/network/ifcfg-'+virtual_host_interface %}
```
smpsgill: same comment as before: ~ instead of +

smpsgill (approved):
looks good! just a minor comment

netweaver/ha_cluster.sls
arbulu89: Yes, totally right! Good catch

arbulu89:
> looks good! just a minor comment

Change force pushed. Thank you @Simranpal

PR #83, opened 2021-05-19
azure-lb resource for ASCS/ERS is not added in the corresponding Reso…

varkoly:
Bug 1186236 - azure-lb resource for ASCS/ERS is not added in the corresponding Resource Group


stefanotorresi (approved):

stefanotorresi:
Thanks Péter!

PR #88, opened 2021-07-05
fix monitoring configuration for ASCS and ERS are not available on all hosts

yeoldegrove:
[SUSE/ha-sap-terraform-deployments/issues/712](https://github.com/SUSE/ha-sap-terraform-deployments/issues/712) describes that there is no monitoring configuration for the ASCS/ERS nodes after a failover.
The pacemaker cluster already handles this correctly but a local config file for the instance on the nodes in `/etc/sap_host_exporter` is missing. Only the instances that are installed get a monitoring configuration at the moment.

This PR fixes this by ...

- creating ASCS|ERS|PAS|AAS monitoring configuration in non-HA use case
- OR create both ASCS and ERS monitoring configuration in HA use case
  - AND deploy PAS exporter on ASCS/PAS host if no dedicated PAS host is used

- not running ASCS and ERS exporter service in HA use case (handled by pacemaker)
  - AND/OR only run PAS exporter in HA use case (if no dedicated PAS host is used)

### DIFF
--- a/netweaver/monitoring.sls
+++ b/netweaver/monitoring.sls
@@ -26,21 +26,26 @@ sap_host_exporter_configuration_{{ exporter_instance }}:
     - require:
       - pkg: prometheus_sap_host_exporter_pkg
 
-# do not run ASCS and ERS exporter service in HA use case (handled by pacemaker)
-# only run PAS exporter in HA use case (if no dedicated PAS host is used)
-{% if not netweaver.ha_enabled or node.sap_instance == 'pas' %}
+# do not enable ASCS and ERS exporter service in HA use case (handled by pacemaker)
+{% if netweaver.ha_enabled and node.sap_instance in ['ascs', 'ers'] %}
+{% set service_status = "disabled" %}
+{% set service_enabled = False %}
+{% else %}
+{% set service_status = "running" %}
+{% set service_enabled = True %}
+{% endif %}
+
 sap_host_exporter_service_{{ exporter_instance }}:
-  service.running:
+  service.{{ service_status }}:
     - name: prometheus-sap_host_exporter@{{ exporter_instance }}
-    - enable: True
+    - enable: {{ service_enabled }}
     - restart: True
     - require:
-      - netweaver_install_{{ instance_name }}
       - pkg: prometheus_sap_host_exporter_pkg
       - file: sap_host_exporter_configuration_{{ exporter_instance }}
     - watch:
       - file: sap_host_exporter_configuration_{{ exporter_instance }}
-{% endif %}
 
 {% endif %}
+
 {% endfor %}
--- a/netweaver/monitoring.sls
+++ b/netweaver/monitoring.sls
@@ -6,7 +6,12 @@ prometheus_sap_host_exporter_pkg:
     - name: prometheus-sap_host_exporter
 
 # the sid, instance number pair of a node is unique, so we need to adapt configuration
-{% for node in netweaver.nodes if host == node.host and node.sap_instance != "db" %}
+# do not create configuration for "sap_instance: db"
+{% for node in netweaver.nodes if node.sap_instance != "db" %}
+# create ASCS|ERS|PAS|AAS configuration in non-HA use case
+# OR create both ASCS and ERS configuration in HA use case
+#   AND deploy PAS exporter on ASCS/PAS host if no dedicated PAS host is used
+{% if (not netweaver.ha_enabled and host == node.host) or (netweaver.ha_enabled and (node.sap_instance in ['ascs', 'ers'] or host == node.host)) %}
 {% set sap_instance_nr = '{:0>2}'.format(node.instance) %}
 {% set exporter_instance = '{}_{}{}'.format(node.sid, node.sap_instance.upper(), sap_instance_nr) %}
 {% set instance_name = node.sid~'_'~sap_instance_nr %}
@@ -20,17 +25,22 @@ sap_host_exporter_configuration_{{ exporter_instance }}:
          sap-control-uds: /tmp/.sapstream5{{ sap_instance_nr }}13
     - require:
       - pkg: prometheus_sap_host_exporter_pkg
-      - netweaver_install_{{ instance_name }}
 
+# do not run ASCS and ERS exporter service in HA use case (handled by pacemaker)
+# only run PAS exporter in HA use case (if no dedicated PAS host is used)
+{% if not netweaver.ha_enabled or node.sap_instance == 'pas' %}
 sap_host_exporter_service_{{ exporter_instance }}:
   service.running:
     - name: prometheus-sap_host_exporter@{{ exporter_instance }}
-    - enable: {{ not netweaver.ha_enabled }}
+    - enable: True
     - restart: True
     - require:
+      - netweaver_install_{{ instance_name }}
       - pkg: prometheus_sap_host_exporter_pkg
       - file: sap_host_exporter_configuration_{{ exporter_instance }}
     - watch:
       - file: sap_host_exporter_configuration_{{ exporter_instance }}
+{% endif %}
 
+{% endif %}
 {% endfor %}
--- a/netweaver/monitoring.sls
+++ b/netweaver/monitoring.sls
@@ -6,12 +6,10 @@ prometheus_sap_host_exporter_pkg:
     - name: prometheus-sap_host_exporter
 
 # the sid, instance number pair of a node is unique, so we need to adapt configuration
-# do not create configuration for "sap_instance: db"
-{% for node in netweaver.nodes if node.sap_instance != "db" %}
-# create ASCS|ERS|PAS|AAS configuration in non-HA use case
-# OR create both ASCS and ERS configuration in HA use case
-#   AND deploy PAS exporter on ASCS/PAS host if no dedicated PAS host is used
-{% if (not netweaver.ha_enabled and host == node.host) or (netweaver.ha_enabled and (node.sap_instance in ['ascs', 'ers'] or host == node.host)) %}
+# in non-HA use case create ASCS|ERS|PAS|AAS configuration
+# in HA use case create additional ASCS|ERS configuration on ERS|ASCS
+{% for node in netweaver.nodes if host == node.host or (netweaver.ha_enabled and node.sap_instance in ['ascs', 'ers']) and node.sap_instance != "db" %}
+
 {% set sap_instance_nr = '{:0>2}'.format(node.instance) %}
 {% set exporter_instance = '{}_{}{}'.format(node.sid, node.sap_instance.upper(), sap_instance_nr) %}
 {% set instance_name = node.sid~'_'~sap_instance_nr %}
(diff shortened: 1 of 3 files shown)


netweaver/monitoring.sls
```diff
@@ -20,17 +25,22 @@ sap_host_exporter_configuration_{{ exporter_instance }}:
          sap-control-uds: /tmp/.sapstream5{{ sap_instance_nr }}13
     - require:
       - pkg: prometheus_sap_host_exporter_pkg
-      - netweaver_install_{{ instance_name }}
 
+# do not run ASCS and ERS exporter service in HA use case (handled by pacemaker)
+# only run PAS exporter in HA use case (if no dedicated PAS host is used)
+{% if not netweaver.ha_enabled or node.sap_instance == 'pas' %}
```
arbulu89: This looks pretty risky. We are assuming that the HA cluster formula will be executed later, and with this the exporter resource agent. I wouldn't take this as granted.
I would just prefer to start the exporter with the `enable` as before.
What was the risk of doing that?

Finally, why is the AAS excluded here?

netweaver/monitoring.sls
```diff
@@ -6,7 +6,12 @@ prometheus_sap_host_exporter_pkg:
     - name: prometheus-sap_host_exporter
 
 # the sid, instance number pair of a node is unique, so we need to adapt configuration
-{% for node in netweaver.nodes if host == node.host and node.sap_instance != "db" %}
+# do not create configuration for "sap_instance: db"
+{% for node in netweaver.nodes if node.sap_instance != "db" %}
+# create ASCS|ERS|PAS|AAS configuration in non-HA use case
+# OR create both ASCS and ERS configuration in HA use case
+#   AND deploy PAS exporter on ASCS/PAS host if no dedicated PAS host is used
+{% if (not netweaver.ha_enabled and host == node.host) or (netweaver.ha_enabled and (node.sap_instance in ['ascs', 'ers'] or host == node.host)) %}
```
arbulu89: `node.sap_instance in ['ascs', 'ers']` looks like will create the configuration file for them in all of the nodes, right? Even in the PAS and AAS nodes.
This doesn't look like really desirable (even though it is not harmful).

arbulu89: Do you think we could simply using the next?:
```{% if host == node.host or (netweaver.ha_enabled and node.sap_instance in ['ascs', 'ers']) %}```

We would still have the issue of having the `ascs/ers` configuration files in the `pas/aas`, but well, who cares hehe

**Review by arbulu89 — CHANGES_REQUESTED** [review]
Hi @yeoldegrove ,
I have added some comments (I think the AAS nodes are unhandled)

File: netweaver/monitoring.sls (hunk 2 above)
Comment (yeoldegrove) [reply netweaver/monitoring.sls]: > `node.sap_instance in ['ascs', 'ers']` looks like will create the configuration file for them in all of the nodes, right? Even in the PAS and AAS nodes.
> This doesn't look like really desirable (even though it is not harmful).

I could swear that this worked in my tests... But you're right, it looks suspicious now...

Comment (yeoldegrove) [reply netweaver/monitoring.sls]: > `{% if host == node.host or (netweaver.ha_enabled and node.sap_instance in ['ascs', 'ers']) %}`
> 
> We would still have the issue of having the `ascs/ers` configuration files in the `pas/aas`, but well, who cares hehe

@arbulu89 What do you think of this?
It should make sure ASCS/ERS configs are only deployed on nodes ending with `0(1|2)$` ... 
```
{% if host == node.host or (netweaver.ha_enabled and (host | regex_search('^.*0(1|2)$')) and node.sap_instance in ['ascs', 'ers']) %}
```

File: netweaver/monitoring.sls (hunk 1 above)
Comment (yeoldegrove) [reply netweaver/monitoring.sls]: I wanted to build it the safe way and not interfere with the HA cluster formula (and pacemaker running the systemd resource) at all. In case of a systemd resource this is ok...as pacemaker will just run a probe/monitor and see that the systemd service is already up even if it fails, it will not stop the SAP instance, as it is started later in the group.
How is this "do not touch" approach more risky than running a service that is handled by the cluster?

AAS was excluded as it is supposed to never run on an HA node. It is included in the `if not netweaver.ha_enabled` statement.
But I think now after reviewing it that I got `netweaver.ha_enabled` wrong... This is global and also set on PAS/AAS.

A simple `enable: {{ not netweaver.ha_enabled }}` does disable the service also on PAS and AAS, doesn't it (if ASCS/ERS are HA)? So it will not come up after a reboot.

Another problem is, that `- require: - netweaver_install_{{ instance_name }}` cannot be used this way as ASCS/ERS will not be installed on ERS/ASCS nodes.
Do you think it is okay to remove the requirement at all? Initially I already removed it from the config file section as the config files are needed on both HA nodes.
But than again... one has to consider starting the ASCS/ERS exporter services only on one side and not on both. Also starting exporters on not running instances is bad.
The way to go is most likely just running `netweaver.monitoring` after `netweaver.install_*` which should already be the case.

File: netweaver/monitoring.sls (hunk 2 above)
Comment (yeoldegrove) [reply netweaver/monitoring.sls]: I guess going for hostname matching is always a bad idea...

Comment (yeoldegrove) [reply netweaver/monitoring.sls]: Could one assume the positions in the `netweaver.nodes` array are always `["ascs","ers", "..."]`?
If this would be the case, one could use their positions.

File: netweaver/monitoring.sls (hunk 1 above)
Comment (yeoldegrove) [reply netweaver/monitoring.sls]: This could be a hybrid approach.

```
# do not enable ASCS and ERS exporter service on in HA use case (handled by pacemaker)
{% if netweaver.ha_enabled and node.sap_instance in ['ascs', 'ers'] %}
{% set exporter_running = False %}
{% else %}
{% set exporter_running = True %}
{% endif %}

sap_host_exporter_service_{{ exporter_instance }}:
{% if exporter_running %}
  service.running:
{% else %}
  service.dead:
{% endif %}
    - name: prometheus-sap_host_exporter@{{ exporter_instance }}
    - enable: {{ exporter_running }}
[...]
```

File: netweaver/monitoring.sls (hunk 2 above)
Comment (pirat013) [reply netweaver/monitoring.sls]: Hey,
one additional note here. Thinking about a diskless SBD cluster setup would mean we would have at least three nodes for the ASCS / ERS instances. If it would make sense to keep this feature in mind because such a deployment would be really interesting for the CSP's. Because they get rid of the SBD device as stonith method in that case. I guess this could have an impact for the solution you looking for at the moment.

Comment (arbulu89) [reply netweaver/monitoring.sls]: I wouldn't take this assumption of using neither the hostname nor the position of the sap instances

@pirat013, I guess you mean that the 3rd node would be a quorum node, right? This formula doesn't really care of this, as the ASCS or ERS shouldn't go to this node

File: netweaver/monitoring.sls (hunk 1 above)
Comment (arbulu89) [reply netweaver/monitoring.sls]: This looks pretty much the same we do in the `saphanabootstrap-formula`: https://github.com/SUSE/saphanabootstrap-formula/blob/master/hana/monitoring.sls#L74

Ok, let's go with this

File: netweaver/monitoring.sls (hunk 2 above)
Comment (arbulu89) [reply netweaver/monitoring.sls]: @yeoldegrove If we want to be purist and have the ASCS/ERS configuration files only on those nodes, we could even have 2 loops, the first one with 
```
{% for node in netweaver.nodes if host == node.host and node.sap_instance != "db" %}
```

And the 2nd only executed with `ascs/ers` to find if there is any other `ascs/ers` to ship the config file to them too.
It looks costly, but maybe we need to do it

**Review by arbulu89 — COMMENTED** [review]
@yeoldegrove Added some more comments

File: netweaver/monitoring.sls (hunk 2 above)
Comment (yeoldegrove) [reply netweaver/monitoring.sls]: @arbulu89 I finally came up with something in 68286ae that is to my liking ;)
It does not need a 2nd loop but solves everything with a combination of `{% if .. %}` and `- onlyif:`.

File: netweaver/monitoring.sls (hunk 1 above)
Comment (yeoldegrove) [reply netweaver/monitoring.sls]: @arbulu89 fixed in 4ffafae

File: netweaver/monitoring.sls [hunk 3]
```diff
@@ -20,17 +23,37 @@ sap_host_exporter_configuration_{{ exporter_instance }}:
          sap-control-uds: /tmp/.sapstream5{{ sap_instance_nr }}13
     - require:
       - pkg: prometheus_sap_host_exporter_pkg
-      - netweaver_install_{{ instance_name }}
+# on HA use case deploy ASCS|ERS on ERS|ASCS
+{% if netweaver.ha_enabled and node.sap_instance in ["ascs", "ers"] %}
+    - onlyif:
+      - test -d /usr/sap/{{ node.sid }}/ASCS* || test -d /usr/sap/{{ node.sid }}/ERS*
```
arbulu89: I would put maybe here some `node.sid.upper()` (or the method to uppercase the string) just in case

netweaver/monitoring.sls
```diff
@@ -20,17 +23,37 @@ sap_host_exporter_configuration_{{ exporter_instance }}:
...
+{% endif %}
+
+# do not enable ASCS and ERS exporter service in HA use case (handled by pacemaker)
+{% if netweaver.ha_enabled and node.sap_instance in ['ascs', 'ers'] %}
```
arbulu89: If we store the output of this condition, maybe we could put these values as one lines in the state itself

arbulu89 (changes requested):
Hi @yeoldegrove ,
Really promising! Just one observation. I see the condition `if netweaver.ha_enabled and node.sap_instance in ['ascs', 'ers']` repeated in many places. Could be store the output of this condition in some variable and use it in all the places? It would reduce a lot of duplicated code and make things easier if we need to change something.

netweaver/monitoring.sls
yeoldegrove: @arbulu89 fixed in fbc9d8e

netweaver/monitoring.sls
yeoldegrove: @arbulu89 refactored this in 15eecaf

arbulu89 (approved):
Nice! Thank you

yeoldegrove:
I will do some final testing and than merge.

netweaver/monitoring.sls
pirat013: @arbulu89 the third node which I had in mind is a valid SAP node which could run the ASCS or ERS. Such a scenario has many advantages for customer and the CSP setups. This node is not limited to be only a quorum node.

arbulu89: Ok, so if this is the case I'm afraid that the monitoring is the last thing that worries me, because most probably the rest of the deployment would fail and we would need to do major changes

pirat013: You are right and I agree. 
If we did not plan to support such a scenario we do not need to take care for this circumstance for the monitoring feature. My point was just keep in mind that we have a supported solution which is interesting for Cloud setups (the request came initially from Azure) and is based on three or five or more nodes. Initially the ASCS and ERS are only installed / available on two nodes but could run on all the other nodes as well.

PR #90, opened 2021-08-27
fix non-existing procedure file when running NW import on a HANA scale-out system

yeoldegrove:
`sapinst` for DB import complains about missing `HdbLandscapeReorgCheckProcedure.SQL` file on scale-out deployments.
The `DONOTUSEFILE` parameter prevents using this file and the import works.

### DIFF
--- a/sapnwbootstrap-formula.changes
+++ b/sapnwbootstrap-formula.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug 27 08:19:27 UTC 2021 - Eike Waldt <waldt@b1-systems.de>
+
+- Version bump 0.6.7
+  * fix non-existing procedure file when running NW import on
+    a HANA scale-out system
+
+-------------------------------------------------------------------
 Thu Jul  7 13:32:06 UTC 2021 - Eike Waldt <waldt@b1-systems.de>
 
 - Version bump 0.6.6
--- a/templates/db.inifile.params.j2
+++ b/templates/db.inifile.params.j2
@@ -181,7 +181,8 @@ NW_readProfileDir.profileDir = {{ sapmnt_path }}/{{ sid.upper() }}/profile/
 archives.downloadBasket = {{ download_basket }}
 
 # Specify whether a Landscape Reorg Check Procedures file 'HdbLandscapeReorgCheckProcedure.SQL' is to be used. Possible values are 'USEFILE', 'DONOTUSEFILE','DONOTRUN'. Default value is 'USEFILE'.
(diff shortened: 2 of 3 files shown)


templates/db.inifile.params.j2
```diff
@@ -181,7 +181,7 @@ NW_readProfileDir.profileDir = {{ sapmnt_path }}/{{ sid.upper() }}/profile/
 archives.downloadBasket = {{ download_basket }}
 
 # Specify whether a Landscape Reorg Check Procedures file 'HdbLandscapeReorgCheckProcedure.SQL' is to be used. Possible values are 'USEFILE', 'DONOTUSEFILE','DONOTRUN'. Default value is 'USEFILE'.
-# hanadb.landscape.reorg.useCheckProcedureFile = USEFILE
+hanadb.landscape.reorg.useCheckProcedureFile = DONOTUSEFILE
```
arbulu89, line 185: How did you find out this?
Is there any guide that you followed?
Just to make sure that we don't create any other inconvenience with this change

yeoldegrove: When not setting this, there was the error message that `HdbLandscapeReorgCheckProcedure.SQL` could not be found.
Looking at the comments of the ini file it was quite obvious what to change to fix it.
There is also a similar line already existing: https://github.com/SUSE/sapnwbootstrap-formula/blob/cf7d0620a6841059b423231d2c4aad7646a9e480/templates/db.inifile.params.j2#L187

Sadly I could not fine any documentation about this parameter.

arbulu89: Ok, I get it.
I was just wondering if we could put the error message somewhere, to make it even more explicit what this change is fixing. Maybe as a comment in this exact file, on top of the change value
In any case, let's move forward with the solution

sapnwbootstrap-formula.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Aug 27 08:19:27 UTC 2021 - Eike Waldt <waldt@b1-systems.de>
+
+- Version bump 0.6.7
+  * fix non-existing procedure file when running NW import on a HANA scale-out system
```
arbulu89: I would split this entry in 2 lines to make the text more compacted as the other entries in the file

arbulu89 (approved):

sapnwbootstrap-formula.changes
yeoldegrove: fixed

templates/db.inifile.params.j2
yeoldegrove: Added an additional comment right above the parameter.

PR #92, opened 2022-01-12
enable S/4HANA 2020 and 2021

yeoldegrove:
Search for "S4HANA" string instead of naming all S/4HANA versions.
* enables S/4HANA 2020
* enables S/4HANA 2021

### DIFF
--- a/templates/db.inifile.params.j2
+++ b/templates/db.inifile.params.j2
@@ -34,7 +34,7 @@ HDB_Schema_Check_Dialogs.schemaPassword = {{ schema_password }}
 # HDB_Schema_Check_Dialogs.validateSchemaName = true
 
 # Use SAP HANA Media on CD, do not ask for SAR archives
-{%- if product_id_header in ['S4HANA1809', 'S4HANA1909'] %}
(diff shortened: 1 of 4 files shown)


templates/db.inifile.params.j2
```diff
@@ -104,7 +104,7 @@ NW_GetMasterPassword.masterPwd = {{ master_password }}
 # NW_HDB_DB.javaSchemaPassword =
 
 # Install the SAP HANA database client in a central or local directory. Possible values are: 'SAPCPE', 'LOCAL'
-{%- if product_id_header in ['S4HANA1809', 'S4HANA1909'] %}
+{%- if 'S4HANA' in product_id_header %}
```
arbulu89: This comparisons were done to avoid the usage of `S4HANA1709` in there, as the installation failed in this version.
Keep this in mind, as we don't want new installations updates breaking older version

yeoldegrove:
@arbulu89 I changed the code to not break the old behavior.
In the meantime I also tested it with S/4HANA 2020 in GCP and it works after adapting the https://github.com/SUSE/ha-sap-terraform-deployments code a bit https://github.com/SUSE/ha-sap-terraform-deployments/pull/808

arbulu89 (approved):

PR #95, opened 2022-02-08
backup HANA after initial DB import (if SR is enabled)

yeoldegrove:
This adds the feature of creating a HANA DB backup after the initial "Netweaver / S/4HANA import" is done (if HANA SR is enabled).
It uses a new pillar switch called `netweaver.hana.ha_enabled` for this, which can later be set via a grain.

After the very large initial "Netweaver / S/4HANA import", the 1st backup on the HANA (done by `saphanabootstrap-formula`) differs to much to re-initiate HANA System Replication. This is why this new backup is needed.

## helps to fix
https://github.com/SUSE/ha-sap-terraform-deployments/issues/739
https://github.com/SUSE/ha-sap-terraform-deployments/issues/803

## prerequisites
`salt-shaptools-0.3.14` including these fixes:
https://github.com/SUSE/salt-shaptools/pull/84
https://github.com/SUSE/salt-shaptools/pull/86

### DIFF
--- a/netweaver/install_db.sls
+++ b/netweaver/install_db.sls
@@ -92,4 +92,17 @@ remove_db_inifile_{{ instance_name }}:
     - require:
       - create_db_inifile_{{ instance_name }}
 
+{%- if netweaver.hana.ha_enabled|default(false) is sameas true %}
+backup_db_{{ netweaver.hana.host }}_{{ netweaver.hana.sid }}_{{ hana_instance }}:
+  module.run:
+    - hana.query:
+      - host: {{ netweaver.hana.host }}
+      - port: 3{{ hana_instance }}13
+      - user: SYSTEM
+      - password: {{ netweaver.hana.password }}
+      - query: "BACKUP DATA FOR FULL SYSTEM USING FILE ('post_db_import')"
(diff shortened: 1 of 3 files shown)


melzer-b1 (approved):
looks ok

melzer-b1 (approved):
looks ok

netweaver/install_db.sls
```diff
@@ -92,4 +92,17 @@ remove_db_inifile_{{ instance_name }}:
     - require:
       - create_db_inifile_{{ instance_name }}
 
+{%- if netweaver.hana.ha_enabled|default(false) is sameas true %}
```
arbulu89: I don't like the `ha_enabled` parameter.
I would rewrite it to `sr_enabled`

arbulu89 (approved):
Marking the PR as green.
But I think there might be some other better way of knowing if a new backup is needed or not.
This `ha_enabled` thing looks something that might be true, but we just know it by experience

netweaver/install_db.sls
yeoldegrove: done

